PR #3, opened 2015-03-06
Correct always forgotten PVR_CLIENT_VERSION

hudokkow:
Haven't looked at the code so don't know if this is absolutely necessary.
@afedchin?


afedchin:
It doesn't necessary because it used in GetBackendVersion, iptvsimple doesn't have a backend so this method simply returns API version simply for info.

opdenkamp:
why don't you just use `XBMC_PVR_API_VERSION`?

afedchin:
You are right. done in master.

PR #6, opened 2015-03-20
v1.11.0 + API 1.9.6 changes

afedchin:
see title.

PR #7, opened 2015-03-21
Fix linux build after #6

afedchin:
@wsnipex can you check linux build with this please?


wsnipex:
yes, that works fine, thanks

afedchin:
thank you too

PR #22, opened 2015-06-17
Live stream pause feature

BytEvil:


afedchin:
this only enables pause button. In general this will not work. after a long pause user will not be able to resume playback because stream will be closed by timeout. To properly implement pause and timeshift addon should handle streams instead of delegating this to the Kodi.

PR #24, opened 2015-06-30
Bump add-on (libplatform dependency)

hudokkow:


afedchin:
I don't understand this. Just bump version to fix platform dependency?

hudokkow:
Yes, statically linked against libplatform

Jalle19:
@afedchin without bumping the version number Kodi won't update the addon.

afedchin:
perhaps this is a linux specific thing. As I know pvr addons distributed with Kodi together.

afedchin:
sorry for noise

PR #43, opened 2015-09-15
PVR API 4.1.0 version bump

metaron-uk:
see https://github.com/xbmc/xbmc/pull/8032


ksooo:
Will come up with a complete PR for the API bump, including Addon version bump and changelog.

PR #53, opened 2015-12-28
kodi.guilib is mandatory, enforce it

hudokkow:


ksooo:
Please do not merge (yet).

pvr.iptvsimple does not depend on guilib and guilib is only "mandatory" due to a conceptional bug in the PVR addon API, imo. Refer to https://github.com/kodi-pvr/pvr.hts/pull/153#issuecomment-167517508 ff for details.

PR #73, opened 2016-04-14
Change to Add-on API 2 library type

AlwinEsch:
This are the changes for the add-on to use API level 2 system.

The rework for Kodi is visible here: https://github.com/xbmc/xbmc/pull/9368

Add-on tested here on Linux and works OK.


Jalle19:
@AlwinEsch is this supposed to be for Krypton?

PR #75, opened 2016-04-22
adapt to PVR API 5.2.0

janbar:


janbar:
related to xbmc/xbmc#9662

PR #83, opened 2016-06-12
Increase logging, parse M3U files lacking EXTM3U marker.

dw:
Needed to handle reading directly M3U from some dodgy IPTV providers.


Jalle19:
Can you rebase and force push to trigger Travis CI again? It seems this PR was forgotten. The changes look and sound reasonable to me - while I don't generally like the idea of catering to people who fail at their core business (IPTV provider providing broken playlists) the changes are fairly harmless.

PR #84, opened 2016-06-18
Language zh_TW updated

wstephen:
Tested and found working in Windows 10 Jarvis 16.1
pvr iptv simple client


MartijnKaijser:
Strings are all handled through Transifex

wstephen:
Hi, Thanks

Sorry, I am not aware of that.  I have just signed with Transifex in the name of wkkstephen.  I found there is also a Martijin (Martijn Kaijser) as a project manager.  I tried to locate “pvr.iptvsimple” in order to help improving the zh_TW translation.

I guess this isn’t an endorsed way, isn’t it.  I mean a volunteer can not initiate an improvement, am I understanding it correctly?

Best Regards!


Stephen

PR #90, opened 2016-10-11
fixup build on windows after 053298623e8a58153eab425d1f2c7d892bc15ac2

afedchin:


src/PVRIptvData.cpp
```diff
@@ -764,14 +764,20 @@ int PVRIptvData::ParseDateTime(std::string& strDate, bool iDateFormat)
...
-  long offset = std::localtime(&current_time)->tm_gmtoff;
+  long offset = 0;
+#ifndef TARGET_WINDOWS
+  offset = -std::localtime(&current_time)->tm_gmtoff;
+#else
+  _get_timezone(&offset);
+#endif // TARGET_WINDOWS
   
   long offset_of_date = (hours * 60 * 60) + (minutes * 60);
-  if (sign == '-') {
```
Jalle19: Why did you change the logic?

afedchin: I didn't change the logic. I've changed sign where it's required (see above at L769 and below L780) because _get_timezone it's actually inverse version of tm_gmtoff.

Jalle19: You're totally right.

Issue #31, opened 2015-07-20
channel number


rasmu:
Hello,
I had created a topic on kodi forum a while with no solution (http://forum.kodi.tv/showthread.php?tid=208312), so I try here.

The addon generate a channel number like this : first channel in list is number 1, etc....
So you input "5" on your remote and the addon open the 5th channel in the list.

Sometimes, people know the channel numbers by heart (on your DSL box or DVB-T).
For example, BeinSport1 is number 46 on my DSL box but on the addon, it's 35 because it's the 35th channel on the list.

In the iptv m3u format, there's no tag to set the channel number.
Maybe in the channel name we could insert the channel number between two separators, I don't know.

What do you think ?
Thanks for your feedback.

hudokkow:
Same thing? -> http://forum.kodi.tv/showthread.php?tid=230226

rasmu:
No, my problem is just with iptvsimple client not with a backend.
What I had understood is : iptvsimple is not a backend so you can't specify a channel number in Kodi.

Forst:
You could as well name the channels as
```
01 Name
02 Name
```
and so on.

afedchin:
fixed by #92

PR #92, opened 2016-11-24
add channel number tag

excorp:
Added channel number tag to m3u file.

It is useful for syncing channel number with backend in kodi.


excorp:
I found something similar in issue #31

src/PVRIptvData.cpp
```diff
@@ -401,6 +404,11 @@ bool PVRIptvData::LoadPlayList(void)
           strTvgLogo = strChnlName;
         }
 
+        if (!strChnlNo.empty()) 
+	{
```
Jalle19: Whitespace issue

excorp: Thank you for answer. I do not know because I'm a beginner, do I have to modify the source and commit again?

PR #102, opened 2017-01-27
Updated Line Length

Shani-08:
Urls could be longer than 1024.

### DIFF
--- a/src/PVRIptvData.cpp
+++ b/src/PVRIptvData.cpp
@@ -327,8 +327,8 @@ bool PVRIptvData::LoadPlayList(void)
   tmpChannel.strTvgLogo     = "";
   tmpChannel.iTvgShift      = 0;
 
-  char szLine[1024];
-  while(stream.getline(szLine, 1024))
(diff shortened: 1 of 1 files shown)


Jalle19:
Do you have an example of this?

src/PVRIptvData.cpp
```diff
@@ -327,8 +327,8 @@ bool PVRIptvData::LoadPlayList(void)
   tmpChannel.strTvgLogo     = "";
   tmpChannel.iTvgShift      = 0;
 
-  char szLine[1024];
-  while(stream.getline(szLine, 1024))
+  char szLine[2000];
```
afedchin: max length of GET request is 4kB, would it make sense to increase to 4096?

Shani-08: yep, that will do too.

afedchin:
please squash the commits.

Shani-08:
OK i made a mess of it while squashomg the commits. i will do another PR.

PR #104, opened 2017-01-31
Increase Line Length to standard 4k

Shani-08:
Some Urls are bigger than 1k length, now using standard.
Patch from Master.


Jalle19:
Thanks a lot!

PR #108, opened 2017-02-09
Android tv - Google tv 

superdevil669:
Hello I would like to know how to install your version of PVR Iptvsimple on an android box. If it would be possible to create a small wiki. Thank you


afedchin:
recordings will never be a part of this addon.

PR #117, opened 2017-04-11
Support multiple groups per channel

mttronc:
This PR adds support for multiple groups per channel, separated by semicolon.

Example:
`group-title="EN;IPTV;News"`

### DIFF
--- a/src/PVRIptvData.cpp
+++ b/src/PVRIptvData.cpp
@@ -316,9 +316,9 @@ bool PVRIptvData::LoadPlayList(void)
 
   int iChannelIndex     = 0;
   int iUniqueGroupId    = 0;
-  int iCurrentGroupId   = 0;
   int iChannelNum       = g_iStartNumber;
   int iEPGTimeShift     = 0;
+  std::vector<int> iCurrentGroupId;
 
   PVRIptvChannel tmpChannel;
   tmpChannel.strTvgId       = "";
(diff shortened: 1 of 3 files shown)


Jalle19:
Can you also update the version number in addon.xml.in and add a corresponding changelog entry?

src/PVRIptvData.cpp
```diff
@@ -468,10 +475,11 @@ bool PVRIptvData::LoadPlayList(void)
 
       iChannelNum++;
 
-      if (iCurrentGroupId > 0)
+      std::vector<int>::iterator it;
+      for (it = iCurrentGroupId.begin(); it != iCurrentGroupId.end(); ++it)
```
afedchin: use `auto` for `it`,
```cpp
for (auto it = iCurrentGroupId.begin(); it != iCurrentGroupId.end(); ++it)
```

src/PVRIptvData.cpp
```diff
@@ -468,10 +475,11 @@ bool PVRIptvData::LoadPlayList(void)
...
       {
-        channel.bRadio = m_groups.at(iCurrentGroupId - 1).bRadio;
-        m_groups.at(iCurrentGroupId - 1).members.push_back(iChannelIndex);
+        channel.bRadio = m_groups.at(*it - 1).bRadio;
```
afedchin: I'm not sure that `*it - 1` will not point to nirvana if there is only one group.

mttronc: This is fine.
The lowest possible value for an int in `iCurrentGroupId` is `1` because:

`int iUniqueGroupId    = 0;`
`group.iGroupId = ++iUniqueGroupId;`
`iCurrentGroupId.push_back(iUniqueGroupId);`

afedchin: got it

src/PVRIptvData.cpp
```diff
@@ -428,22 +429,28 @@ bool PVRIptvData::LoadPlayList(void)
 
         if (!strGroupName.empty())
         {
-          strGroupName = XBMC->UnknownToUTF8(strGroupName.c_str());
+          std::stringstream stream(strGroupName);
```
mttronc: This overwrites `std::stringstream stream` from line 312. Gonna rename it to `streamGroup`

mttronc:
If I do a force push for the changes, will it break this PR or will it be fine?

I also want to note that I casually fixed a bug:
If a channel has no group defined, its group is set to the previous channel's group. This happens due to `iCurrentGroupId` not being reset every loop.
I will mention this in the changelog as well.

Also I am looking forward to backport this PR to Krypton once it's all fine and merged.

afedchin:
> If I do a force push for the changes, will it break this PR or will it be fine?

it will be fine.

> If a channel has no group defined, its group is set to the previous channel's group. 

please don't touch this. It's a desired behavior.

mttronc:
Okay, I will restore this behaviour.

Jalle19:
@mttronc can you rebase?

asciidisco:
@Jalle19 Might be a stupid question, but I´m a bit confused with all the PVR changes (and what gets backported, what not...), so, will this feature land in the Version for Kodi 17 as well as in 18, or is it a 18 only feature?

Jalle19:
For now this will only be included in v18. Nothing prevents someone from backporting the changes to the v17 branch though.

asciidisco:
@Jalle19 Thank you for clarification.

PR #121, opened 2017-04-29
change to automatic interface version set

AlwinEsch:
Related to Kodi Pull Request xbmc/xbmc#12029 to set interface versions automatic.


AlwinEsch:
Not so fast @afedchin :-D want to have a OK first about them. Is nice for me to set the versions automatic, but must be accepted.

Can remove a lot of work in future where then only a recompile is needed, but a bit hard to understand in History.

afedchin:
sorry, I didn't pay attention to repo. I thought that it's my own :(

PR #137, opened 2017-07-11
Pause button

petto10:
Can you add the pause button?
Thank You


afedchin:
1. pause will never be added. 
2. your PR isn't related to your request.

PR #141, opened 2017-08-05
Add PVR Addon API 5.5.0 support

sunsettrack4:


ksooo:
Thanks. But if you want to make the addon *working* again (not just compile and link), you need to properly implement  `GetLiveStreamURL`. Basically, it should return what the old implementation put into `PVR_CHANNEL::strStreamURL`. As `GetLiveStreamURL` gets called whenever a channel stream shall be opened, this gives you more flexibility what URL to return, compared to the old way filling `PVR_CHANNEL::strStreamURL` in `GetChannels`, becuase the latter is generally only called once during Kodi lifetime.

rinhudio:
@ksooo is it that hard to implement the new api? I thought with this new pvr api it is much simpler?!

Isn't it just 10min work for one of you kodi core guys to fix this addon? (I have not really a feeling for the complexity of the required changes...)

ksooo:
Problem is that team Kodi has very limited development resources. We do, what we can but actively maintaining all pvr add-ons is nothing we are able to do, currently. Sorry, wet need support from the community.

rinhudio:
If the development resources are limited and team kodi is unable to handle the consequence of changing the core api - why did you do that. It is a shame to break a core feature and then to say it is the community's task to fix it!
For python addons the communty is much larger then the binary community, since not everyone is able to write c++.

Of course nobody likes to apply the changes to 15+ pvr addons. But just cherry picking the favorite tasks and leaving the rest for the community is...

If you agree that pvr is a core feature of kodi, then the core team should feel responsible at least for the minmal maintenance to keep the plugins working (not to implement new features).

@sunsettrack4 as part of the community, do you feel able to implement the new api as described by ksooo?

peak3d:
@rinhudio any reason why you are using a development branch?
IIRC pvr.iptvsimple works well on the released version K17

ksooo:
If you do not like what we do and you feel like we are a******s, you can choose another media player / pvr software whenever you want.

Jalle19:
@rinhudio this is alpha software, surely someone will fix the addon long before the next version of Kodi is released

ksooo:
Well, I did the monkey work to keep all pvr add-ons alive for over 2 years. And the first time I decided that I do not have enough time for that, because the changes require insights into every single add-on implementation, almost immediately some [moderated] shows up and starts crying. I'm so sick about you [moderated] leechers! OTOH, same people cry "more features, more features, why does PiP not work" and the like. Did I say how sick I'm about this?

rinhudio:
Nobody said that you are a******s. That was misinterpreted. I noticed that you, @ksooo, keeped the pvr addons synced for a long time. I'm thankful for that and also for the work of all the other contributors.

@Jalle19 this is not alpha software but the development branch. It is not tagged as alpha and under heavy development. I'm aware of that. I don't use it for productive but on a second device to follow the development and to check my addons to be ready before a new version is released.
But it doesn't matter why I use it. I was excited because I read that there is nobody making the pvr addons ready for the new api - and that the next release will probability be shipped without some pvr addons. This gave the impression, that the core was improved/cleaned up (that is praiseworthy), but that leads to a huge lost of functionality.
However, this is not the right place for such a discussion. I'm sorry for hijacking this PR.

Jalle19:
@rinhudio like I said, don't worry, someone will fix it sooner or later

ksooo:
Your post was offering and demanding - period. Nice words afterwords don't help. Think at least twice before you write bullshit like "its a shame" and the like to the people who deliver you free software, spending their free time, without getting a single buck for this.

PR #147, opened 2017-08-07
[win10] do not include StdString header (not used actually)

afedchin:
see title


Jalle19:
@afedchin can you fix the build failures?

afedchin:
I'm trying, but with not success. I can't understand what's wrong with `stat64`

afedchin:
@Jalle19 fixed.

PR #153, opened 2017-08-16
Backport: Added support for multiple groups per channel, separated by…

flubshi:
… semicolon

It is a backport of: #117 

All credits to @mttronc 

Backport was requested within the discussion of the original PR and here, to make maintance of a public channel list easier (german): https://github.com/jnk22/kodinerds-iptv/issues/47

[Testing]
I compiled and runtime tested it with Kodi 17.3 on linux. For me, with this patch, it is possible to assign multiple groups to a single channel (semicolon separated) as with the v18 patch.

### DIFF
--- a/pvr.iptvsimple/addon.xml.in
+++ b/pvr.iptvsimple/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.iptvsimple"
-  version="2.4.13"
+  version="2.4.14"
   name="PVR IPTV Simple Client"
   provider-name="nightik">
   <requires>
--- a/pvr.iptvsimple/changelog.txt
+++ b/pvr.iptvsimple/changelog.txt
@@ -1,3 +1,6 @@
+v2.4.14
+- Backport: Added support for multiple groups per channel, separated by semicolon
+
 v2.4.13
 - fixed install path in the Debian package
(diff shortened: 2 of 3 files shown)


Jalle19:
This is how it should be done, good job!

MartijnKaijser:
This wont be shipped if our repo reference isn't bumped as well. Same for any other addon btw

Jalle19:
@MartijnKaijser I thought that was all automatic these days? Can you point me at the right repository?

MartijnKaijser:
@Jalle19 yes for Kodi master we use addon master but for release branches we use fixed tags to prevent random crap/accidents happening.
https://github.com/xbmc/repo-binary-addons/tree/Krypton

Jalle19:
We should perhaps consider changing that. If an addon developer screws up a release it shouldn't be our problem.

MartijnKaijser:
That's an added burden on release time i don't want to deal with cause you would have to check if all compiles again/still. If not you can't release cause it's missing an addon and users start yelling at us.
Solution is binary repo so we don't ship at all and they have to install like python addons. 
Also having them handle the bump keeps them in the loop just like we do for python addons as the dev is responsible for having an up to date version available.
At least that's my view on it.

flubshi:
@MartijnKaijser or @Jalle19 maybe one of you could tag a new version and bump the binary reference? Maybe before https://github.com/xbmc/xbmc/pull/12698 ?

PR #169, opened 2017-12-21
Fix multithreading issues

a1rwulf:
Public methods that are part of the addon interface maybe called from multiple threads.
For example json rpc.


a1rwulf:
@afedchin maybe you want to have a short look, should be quite straight forward.

ksooo:
Please add Addon version bump and changelog update.

a1rwulf:
Done.

ksooo:
Thanks.

PR #174, opened 2018-01-15
added: pass on kodi properties from m3u

notspiff:
in particular this allows for inputstream.adaptive streams (dash etc)

@ksooo @peak3d


notspiff:
added the bump.

ksooo:
Thanks, it would be nice if you update https://github.com/kodi-pvr/pvr.iptvsimple/blob/master/pvr.iptvsimple/changelog.txt as well.

notspiff:
and the changelog.

ksooo:
Thansk @notspiff

PR #188, opened 2018-06-20
PVR API 5.10.1

ksooo:
For details see https://github.com/xbmc/xbmc/pull/14086

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -338,14 +338,16 @@ PVR_ERROR GetChannelStreamProperties(const PVR_CHANNEL* channel, PVR_NAMED_VALUE
   if (!channel || !properties || !iPropertiesCount)
     return PVR_ERROR_SERVER_ERROR;
 
-  if (*iPropertiesCount < 1)
+  if (*iPropertiesCount < 2)
     return PVR_ERROR_INVALID_PARAMETERS;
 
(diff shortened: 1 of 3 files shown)


src/client.cpp
```diff
@@ -345,7 +345,9 @@ PVR_ERROR GetChannelStreamProperties(const PVR_CHANNEL* channel, PVR_NAMED_VALUE
   {
     strncpy(properties[0].strName, PVR_STREAM_PROPERTY_STREAMURL, sizeof(properties[0].strName) - 1);
     strncpy(properties[0].strValue, m_currentChannel.strStreamURL.c_str(), sizeof(properties[0].strValue) - 1);
-    *iPropertiesCount = 1;
+    strncpy(properties[1].strName, PVR_STREAM_PROPERTY_ISREALTIMESTREAM, sizeof(properties[1].strName) - 1);
+    strncpy(properties[1].strValue, "true", sizeof(properties[1].strValue) - 1);
+    *iPropertiesCount = 2;
```
palinek: Shouldn't be also the above check:
```c++
if (*iPropertiesCount < 1)
  return PVR_ERROR_INVALID_PARAMETERS;
```
changed to `if (*iPropertiesCount < 2)` ?

ksooo: Yes. Good catch. Thank you.

ksooo: Fixed.

PR #189, opened 2018-06-25
Add support for #EXT-X-PLAYLIST-TYPE in M3U

a1rwulf:
If we find a line #EXT-X-PLAYLIST-TYPE:VOD somewhere
before the stream definition, we do not flag the
item as realtime.
If nothing is stated or the line is #EXT-X-PLAYLIST-TYPE:EVENT
we flag the channel as realtime.

### DIFF
--- a/src/PVRIptvData.cpp
+++ b/src/PVRIptvData.cpp
@@ -45,6 +45,7 @@
 #define GROUP_NAME_MARKER       "group-title="
 #define KODIPROP_MARKER         "#KODIPROP:"
 #define RADIO_MARKER            "radio="
+#define PLAYLIST_TYPE_MARKER    "#EXT-X-PLAYLIST-TYPE:"
 #define CHANNEL_LOGO_EXTENSION  ".png"
 #define SECONDS_IN_DAY          86400
 #define GENRES_MAP_FILENAME     "genres.xml"
@@ -318,7 +319,7 @@ bool PVRIptvData::LoadPlayList(void)
 
   /* load channels */
   bool bFirst = true;
-
+  bool bIsRealTime  = true;
   int iChannelIndex     = 0;
   int iUniqueGroupId    = 0;
   int iChannelNum       = g_iStartNumber;
@@ -352,7 +353,7 @@ bool PVRIptvData::LoadPlayList(void)
       {
         strLine.erase(0, 3);
       }
-      if (StringUtils::Left(strLine, (int)strlen(M3U_START_MARKER)) == M3U_START_MARKER)
+      if (StringUtils::Left(strLine, strlen(M3U_START_MARKER)) == M3U_START_MARKER)
       {
         double fTvgShift = atof(ReadMarkerValue(strLine, TVG_INFO_SHIFT_MARKER).c_str());
         iEPGTimeShift = (int) (fTvgShift * 3600.0);
@@ -367,7 +368,7 @@ bool PVRIptvData::LoadPlayList(void)
       }
     }
 
-    if (StringUtils::Left(strLine, (int)strlen(M3U_INFO_MARKER)) == M3U_INFO_MARKER)
+    if (StringUtils::Left(strLine, strlen(M3U_INFO_MARKER)) == M3U_INFO_MARKER)
     {
       bool        bRadio       = false;
       double      fTvgShift    = 0;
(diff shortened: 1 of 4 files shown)


a1rwulf:
@ksooo and @peaked pls review.
I set the livestream flag now if the playlist-type definition is missing.
Not sure what our consense was about the default.

src/PVRIptvData.cpp
```diff
@@ -469,12 +470,23 @@ bool PVRIptvData::LoadPlayList(void)
         tmpChannel.properties.insert({prop, propValue});
       }
     }
+    else if (StringUtils::Left(strLine, (int)strlen(PLAYLIST_TYPE_MARKER)) == PLAYLIST_TYPE_MARKER)
```
ksooo: Please use a c++ style cast here. I see you copied this from some lines above, but please don't copy blindly.

src/PVRIptvData.cpp
```diff
@@ -469,12 +470,23 @@ bool PVRIptvData::LoadPlayList(void)
...
+        tmpChannel.properties.insert({PVR_STREAM_PROPERTY_ISREALTIMESTREAM, "false"});
+    }
     else if (strLine[0] != '#')
     {
       XBMC->Log(LOG_DEBUG,
                 "Found URL: '%s' (current channel name: '%s')",
                 strLine.c_str(), tmpChannel.strChannelName.c_str());
 
+      if (tmpChannel.properties.find(PVR_STREAM_PROPERTY_ISREALTIMESTREAM) == tmpChannel.properties.end())
+        tmpChannel.properties.insert({PVR_STREAM_PROPERTY_ISREALTIMESTREAM, "true"});
```
ksooo: Interesting approach, but couldn't this be coded much simpler, without searching the prop?

src/PVRIptvData.cpp
```diff
@@ -469,12 +470,23 @@ bool PVRIptvData::LoadPlayList(void)
         tmpChannel.properties.insert({prop, propValue});
       }
     }
+    else if (StringUtils::Left(strLine, (int)strlen(PLAYLIST_TYPE_MARKER)) == PLAYLIST_TYPE_MARKER)
+    {
+      std::string value = ReadMarkerValue(strLine, PLAYLIST_TYPE_MARKER);
+      if (value == "EVENT")
+        tmpChannel.properties.insert({PVR_STREAM_PROPERTY_ISREALTIMESTREAM, "true"});
+      else if (value == "VOD")
+        tmpChannel.properties.insert({PVR_STREAM_PROPERTY_ISREALTIMESTREAM, "false"});
```
ksooo: No need to insert the prop if the value is false, as this is the default if the prop is not present.

src/PVRIptvData.cpp
ksooo: I would simple define and fill a bool bIsRealTime and if it is true I would append the prop after the while loop.

ksooo:
Default must be "true".

src/PVRIptvData.cpp
a1rwulf: I had a look at the signature of StringUtils::Left and it actually takes a parameter of type size_t which is exactly what strlen returns.
So the cast is useless at all - will remove it for the other occurrences as well.

a1rwulf:
@ksooo addressed your comments.
You may have a look again thx.

src/PVRIptvData.cpp
```diff
@@ -469,12 +470,21 @@ bool PVRIptvData::LoadPlayList(void)
         tmpChannel.properties.insert({prop, propValue});
       }
     }
+    else if (StringUtils::Left(strLine, strlen(PLAYLIST_TYPE_MARKER)) == PLAYLIST_TYPE_MARKER)
+    {
+      std::string value = ReadMarkerValue(strLine, PLAYLIST_TYPE_MARKER);
+      if (value == "VOD")
```
ksooo: nitpick: you could eliminate the variable here; `if ReadMarkerValue(strLine, PLAYLIST_TYPE_MARKER) == "VOD"` also would do the trick.

a1rwulf: Sure it's a leftover cause I had 2 if's before :)

ksooo:
@peak3d maybe you can give this a runtime-test?

peak3d:
@a1rwulf will test tomorrow, day was long busy

a1rwulf:
Take your time we are not in a hurry ;)

a1rwulf:
ping @peak3d

a1rwulf:
Build failure is unrelated.
Travis file is the same as in other repos.
So I don't see a reason to hold it back.
@ksooo can you merge please - I don't have write permission.

ksooo:
Here we o.

PR #190, opened 2018-06-29
PVRIptvData: Fix considering DST in timestamp conversion

### LINKED ISSUE #158: [Windows] - Wrong EPG Times due to DST (1 Hour Out)
Windows only.
In LibreELEC (Linux), the times are correct.

NZ has just switched to DST so we are now UTC + 13:00.
In Windows, the time is correct and Windows does display right time and is aware it is DST.

However, in KODI - all the EPG programs are out by an hour.
If I set a +1 offset in the settings, they are corrected.

Looks like the issue could be here:
https://github.com/kodi-pvr/pvr.iptvsimple/blob/6e44490841e31e149b1d71348fc29df4a04f89d9/src/PVRIptvData.cpp#L831

**_get_timezone (does not include DST) is used for Windows
tm_gmtoff (includes DST) is used for all other arch.**

It looks like Windows was using tm_gmtoff until the below pull request (October 2016) swapped it to _get_timezone.
https://github.com/kodi-pvr/pvr.iptvsimple/pull/90/files

This may explain why a user reported it's fine in KODI 16:
https://forum.kodi.tv/showthread.php?tid=315206&pid=2618530#pid2618530

TEMP SOLUTION:
1) Go to the PVR IPTV Simple Client Settings and set the Offset to +1.
2) Then go-to the KODI PVR / TV Settings -> Guide -> Clear Data
3) When daylight savings ends, set the offset to 0, and clear the guide data again.

palinek:
1. Don't use the undocumented tm_gmtoff attribute (implementation specific)
2. Unify the logic for all platforms
3. Consider the diff between UTC and local time at the the particular
timestamp

trying to fix #158

### DIFF
--- a/pvr.iptvsimple/addon.xml.in
+++ b/pvr.iptvsimple/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.iptvsimple"
-  version="3.5.7"
+  version="3.5.8"
   name="PVR IPTV Simple Client"
   provider-name="nightik">
   <requires>@ADDON_DEPENDS@</requires>
--- a/pvr.iptvsimple/changelog.txt
+++ b/pvr.iptvsimple/changelog.txt
@@ -1,3 +1,5 @@
+v3.5.8
+- Fix considering DST in timestamp conversion
 
 v3.5.7
 - Correctly show build version in Kodi
--- a/src/PVRIptvData.cpp
+++ b/src/PVRIptvData.cpp
@@ -28,6 +28,7 @@
 #include <sstream>
 #include <stdexcept>
 #include <string>
+#include <iomanip>
 
 #include "zlib.h"
 #include "rapidxml/rapidxml.hpp"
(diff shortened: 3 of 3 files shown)


palinek:
rebased to current master

palinek:
@Rechi @ksooo @afedchin 
Should I rebase this? Or make PR against Matrix? Or is this a NO GO ?

palinek:
rebased to current Leia ...

Sorien:
I tried it from atrifacts but every time I try it epg is empty with no error in logs

I have start and stop timestamps in UTC like `<programme channel="2458735488" start="20190514063000" stop="20190514070000">`

ksooo:
@palinek I think we should go with #248 to fix this. Seems to be simpler (no platform dep etc.) and working as well. What do you think?

palinek:
@ksooo You're free to decide. Actually I don't use iptvsimple and so not suffering from this bug. I just saw this bug back then and wanted to share a solution which we had used in other project.

ksooo:
Superseded  by #257

PR #206, opened 2018-10-31
Fix build version shown in Kodi

### LINKED ISSUE #205: Version string
In client.cpp there is `strBackendVersion = STR(IPTV_VERSION)`. But seems this `IPTV_VERSION` string is not declared anywhere. Should be in Kodi?

vpeter4:
from `Version: IPTV_VERSION`
to `Version: 3.5.5`
fixes #205

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,9 @@ set(IPTV_SOURCES src/client.cpp
 set(IPTV_HEADERS src/client.h
                  src/PVRIptvData.h)
 
+addon_version(pvr.iptvsimple IPTV)
+add_definitions(-DIPTV_VERSION=${IPTV_VERSION})
+
 build_addon(pvr.iptvsimple IPTV DEPLIBS)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -31,4 +31,6 @@ set(IPTV_HEADERS src/client.h
 
 build_addon(pvr.iptvsimple IPTV DEPLIBS)
 
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DIPTV_VERSION=${IPTV_VERSION}")
```
Rechi: Better use `add_definitions(-DIPTV_VERSION=${IPTV_VERSION})` instead of fiddle with `CMAKE_CXX_FLAGS`.
Also it should be move above `build_addon(...)`

vpeter4: Fixed. I need to add call `addon_version()` to actually set the version.

PR #214, opened 2019-01-18
Add support for Timeshifting

phunkyfish:
@ksooo 
I attempted to add Timeshifting support to iptvsimple.

It worked fine with the first stream I tried however not with the second or subsequent due to demuxing. It just hangs on creating the demux I assume because I would have to implement demuxing in the addon.

Happy to do it but not a clue where to start! I looked the the demuxer in TVHeadend but I think that's very specific to hts. Any guidance on how to do this?


phunkyfish:
Sorry, forgot, debug log: https://pastebin.com/BsLYPN8X

phunkyfish:
Nevermind. I read the IPTV archive forum post. Will have a look at the demuxing in myth tv.

PR #216, opened 2019-01-26
[depends] rapidxml: split windows from common

Rechi:

PR #217, opened 2019-02-01
Timeshift

phunkyfish:
v3.6.0
- Added: Support for Timeshifting

As the M3U8 playlist needs to updated on a rolling basis it was not possible to just direct the stream into kodi. Instead download each segment/decrypt etc and then stream into kodi.

Leveraged a C project for the HLS parts (most of it ported): https://github.com/e2iplayer/hlsdl and a concurrent queue for the stream: https://github.com/cameron314/readerwriterqueue


afedchin:
who will support all this code?
my personal as an author of the addon opinion: if you need timeshift or recordings then IPTV addon is not your choice.

phunkyfish:
I'll agree it turned into a bit of a beast. I have no issue with supporting it. It was mentioned on the forum that timeshift was acceptable here but recordings would not be. I thought the community would appreciate it, plus I had some time spare time over the Xmas break. The addon is unchanged unless you enable timeshift.

Ideally something like inputstream.adaptive could handle this but I wasn't sure where to start with that.

phunkyfish:
Strange rapidxml appears to be breaking the build. @Rechi thoughts? Or have I just done something wrong.

Rechi:
#216 fixes the build errors.

phunkyfish:
Great, thank you

CMakeLists.txt
```diff
@@ -11,23 +11,65 @@ find_package(kodiplatform REQUIRED)
 find_package(p8-platform REQUIRED)
 find_package(Rapidxml REQUIRED)
 find_package(ZLIB REQUIRED)
+find_package(OpenSSL REQUIRED)
```
phunkyfish, line 14: @Rechi I used vfs.sftp and vfs.libarchive as a reference but can't see why this fails on nearly every platform. Am I missing something?

Rechi: OpenSSL is missing in depends.

phunkyfish: On Jenkins?

Rechi: in your PR branch

phunkyfish: Oh, now I feel silly. Thanks again for your help.

debian/control
```diff
@@ -2,7 +2,7 @@ Source: kodi-pvr-iptvsimple
 Priority: extra
 Maintainer: Anton Fedchin <anightik@gmail.com>
 Build-Depends: debhelper (>= 9.0.0), cmake, libkodiplatform-dev (>= 16.0.0),
-               kodi-addon-dev, librapidxml-dev, zlib1g-dev
+               kodi-addon-dev, librapidxml-dev, zlib1g-dev, openssl
```
Rechi: openssl is the openssl binary, you need the openssl library and headers (libssl-dev)

phunkyfish: Ah, ok, thanks

ksooo:
> who will support all this code?
my personal as an author of the addon opinion: if you need timeshift or recordings then IPTV addon is not your choice.

@phunkyfish reading this from the addon maintainer I'm not sure it is worth to invest more time into this effort.

phunkyfish:
I just wanted to learn how to get it working. A personal learning exercise 😉

SerhiyRomanov:
Hello @phunkyfish 

> I thought the community would appreciate it

Yes, I appreciate your work because I still don't understand why the timeshift feature doesn't work here. 
My IPTV operator provides an archive for several days and I am very happy about Catchup (Archive) on Matrix, additional thanks for it.

phunkyfish:
I do hope to support timeshift one day.

The new inputstream addon ffmpegdirect that is used for catchup could potentially support it in the future.

phunkyfish:
Closing, timeshift feature released on Matrix using inputstream.ffmpegdirect.

PR #151, opened 2017-08-13
Add support for <sub-title> (episode name) XMLTV tags

Jalle19:
Closes #91

@excorp can you try this?

Issue #91, opened 2016-11-23
EpisodeName Support


kyl416:
Can you add support for EpisodeName via the xmltv file?

The value is in the sub-title field.

I was able to add support by doing the following in PVRIptvData.cpp
```
@@ -274,6 +274,7 @@
     GetNodeValue(pChannelNode, "title", entry.strTitle);
     GetNodeValue(pChannelNode, "desc", entry.strPlot);
     GetNodeValue(pChannelNode, "category", entry.strGenreString);
+    GetNodeValue(pChannelNode, "sub-title", entry.strEpisodeName);
 
     xml_node<> *pIconNode = pChannelNode->first_node("icon");
     if (pIconNode == NULL || !GetAttributeValue(pIconNode, "src", entry.strIconPath))
@@ -715,7 +716,7 @@
       tag.iSeriesNumber       = 0;     /* not supported */
       tag.iEpisodeNumber      = 0;     /* not supported */
       tag.iEpisodePartNumber  = 0;     /* not supported */
-      tag.strEpisodeName      = NULL;  /* not supported */
+      tag.strEpisodeName      = myTag->strEpisodeName.c_str();
       tag.iFlags              = EPG_TAG_FLAG_UNDEFINED;
```

And in PVRIptvData.h
```
@@ -41,6 +41,7 @@
   std::string strPlot;
   std::string strIconPath;
   std::string strGenreString;
+  std::string strEpisodeName;
 };
 
 struct PVRIptvEpgChannel
```

excorp:
After applying the patch as you wrote it, it works well.
Why do not you try a pull request?

kyl416:
I don't really know how to do pull requests

Jalle19:
Google for "how to make a github pull request", there are many good tutorials.

flaviofire:
Hi , 
kyl416  could you compile for windows? I really need this Episode and Seasons....

kyl416:
Any update on getting that pull request merged? This is basically holding me back from upgrading to Leia 18 on my Windows machine since I really don't want to go through the hassle of setting up another Visual Studio just to recompile this one addon like I did for Krypton since it takes several hours to download Visual studio on a slower connection.

Although if there is a more simple process to just compile one addon on windows (and armv8 Android too for that matter), I can't find it documented anywhere.

phunkyfish:
[...]

PR #219, opened 2019-02-11
Add support for <sub-title> (episode name) XMLTV tags

kyl416:
Cleaner version of #151 and hopefully closes my original issue #91


kyl416:
It looks like those checks failed because of an unrelated issue involving rapidxml on some platforms that wiil get cleared up whenever #216 is merged

kyl416:
Now that the rapidxml patch has been merged, is there anyway to have this pull request rechecked for Leia and now Matrix/master?

PR #231, opened 2019-03-23
Update of German terms

CurlingTongs:
- In German, we say "Versatz" instead of "Versetzung" (Usually "Versetzung" is used in a different context. I would even recommend "Zeitverschiebung" or "Zeitunterschied".)
- According to offical German spelling rules, we have to put a "-" between constructed/combined nouns.

### DIFF
--- a/pvr.iptvsimple/resources/language/resource.language.de_de/strings.po
+++ b/pvr.iptvsimple/resources/language/resource.language.de_de/strings.po
@@ -34,11 +34,11 @@ msgstr "Allgemein"
 
 msgctxt "#30011"
 msgid "M3U Play List Path"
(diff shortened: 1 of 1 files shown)


ksooo:
Kodi main language is en_gb, which is maintained through github. All other languages are managed using Transifix. So, language fixes for other languages than en_gb must go through Transifix.

But, as I have the according Transifex privileges (and being native German speaker), I just incorporated the changes you suggest there. Thanks.

CurlingTongs:
Dear Kai,
first, I am a native German speaker as well. Is there any reason for ignoring "our" official German spelling rules? Seems as logical as ignoring the formal traffic rules ("... but not for me!").
Kind regards

PR #242, opened 2019-04-25
[cmake] remove usage of kodi-platform

olafhering:
The pvr.iptvsimple binary addon does not use kodi-platform.
Remove existing references from CMakeLists.txt and debian/control.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -1,7 +1,7 @@
 Source: kodi-pvr-iptvsimple
 Priority: extra
 Maintainer: Anton Fedchin <anightik@gmail.com>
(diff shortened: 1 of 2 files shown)


debian/control
```diff
@@ -1,7 +1,7 @@
 Source: kodi-pvr-iptvsimple
 Priority: extra
 Maintainer: Anton Fedchin <anightik@gmail.com>
-Build-Depends: debhelper (>= 9.0.0), cmake, libkodiplatform-dev (>= 16.0.0),
+Build-Depends: debhelper (>= 9.0.0), cmake, libp8-platform-dev
```
Rechi: please fix the syntax error in this line

olafhering: Done

PR #248, opened 2019-05-13
DST Fix (Matrix)

matthuisman:
This is @buers code.
He is unfamiliar with git so I am making a pull request on his behalf.

Issue with discussion here:
https://github.com/kodi-pvr/pvr.iptvsimple/issues/158

### DIFF
--- a/src/PVRIptvData.cpp
+++ b/src/PVRIptvData.cpp
@@ -53,6 +53,36 @@
 using namespace ADDON;
 using namespace rapidxml;
 
+namespace
+{
+/*
+  Adapted from https://stackoverflow.com/a/31533119
+  Conversion from UTC date to second, signed 64-bit adjustable epoch version.
+  Written by François Grieu, 2015-07-21; public domain. 
+*/
+
+int MakeTime(int year, int month, int day)
+{
+  return year * 365 + year / 4 - year / 100 * 3 / 4 + (month + 2) * 153 / 5 + day;
+}
+
+int GetUTCTime(int year, int mon, int mday, int hour, int min, int sec)
+{
+  int m = mon - 1;
+  int y = year + 100;
+
+  if (m < 2)
+  {
+    m += 12;
+    --y;
+  }
+
+  // compute number of days within constant, assuming appropriate origin
+  return (((MakeTime(y, m, mday) - MakeTime(1970 + 99, 12, 1)) * 24 + hour) * 60 + min) * 60 + sec;
+}
+
+} // unnamed namespace
+
 template<class Ch>
 inline bool GetNodeValue(const xml_node<Ch> * pRootNode, const char* strTag, std::string& strStringValue)
 {
@@ -806,37 +836,27 @@ int PVRIptvData::GetFileContents(std::string& url, std::string &strContent)
 
 int PVRIptvData::ParseDateTime(std::string& strDate, bool iDateFormat)
 {
-  struct tm timeinfo;
-  memset(&timeinfo, 0, sizeof(tm));
+  int year = 2000;
+  int mon = 1;
+  int mday = 1;
+  int hour = 0;
+  int min = 0;
+  int sec = 0;
   char sign = '+';
   int hours = 0;
   int minutes = 0;
 
   if (iDateFormat)
-    sscanf(strDate.c_str(), "%04d%02d%02d%02d%02d%02d %c%02d%02d", &timeinfo.tm_year, &timeinfo.tm_mon, &timeinfo.tm_mday, &timeinfo.tm_hour, &timeinfo.tm_min, &timeinfo.tm_sec, &sign, &hours, &minutes);
+    sscanf(strDate.c_str(), "%04d%02d%02d%02d%02d%02d %c%02d%02d", &year, &mon, &mday, &hour, &min, &sec, &sign, &hours, &minutes);
   else
-    sscanf(strDate.c_str(), "%02d.%02d.%04d%02d:%02d:%02d", &timeinfo.tm_mday, &timeinfo.tm_mon, &timeinfo.tm_year, &timeinfo.tm_hour, &timeinfo.tm_min, &timeinfo.tm_sec);
+    sscanf(strDate.c_str(), "%02d.%02d.%04d%02d:%02d:%02d", &mday, &mon, &year, &hour, &min, &sec);
 
-  timeinfo.tm_mon  -= 1;
-  timeinfo.tm_year -= 1900;
-  timeinfo.tm_isdst = -1;
-
-  std::time_t current_time;
-  std::time(&current_time);
-  long offset = 0;
-#ifndef TARGET_WINDOWS
-  offset = -std::localtime(&current_time)->tm_gmtoff;
-#else
-  _get_timezone(&offset);
-#endif // TARGET_WINDOWS
-
-  long offset_of_date = (hours * 60 * 60) + (minutes * 60);
+  long offset_of_date = (hours * 60 + minutes) * 60;
+  
   if (sign == '-')
-  {
     offset_of_date = -offset_of_date;
-  }
 
-  return mktime(&timeinfo) - offset_of_date - offset;
+  return GetUTCTime(year, mon, mday, hour, min, sec) - offset_of_date;
 }
 
 PVRIptvChannel * PVRIptvData::FindChannel(const std::string &strId, const std::string &strName)


matthuisman:
Test builds here:
https://jenkins.kodi.tv/blue/organizations/jenkins/kodi-pvr%2Fpvr.iptvsimple/detail/PR-248/1/artifacts

buers:
Hi Matt,

 

are you receiving this directly by email? Is it ok to you, to mail with me directly? buers@gmx.de <mailto:buers@gmx.de> .

 

Thanks for your efforts!

 

Von: Matt Huisman <notifications@github.com> 
Gesendet: Dienstag, 14. Mai 2019 00:20
An: kodi-pvr/pvr.iptvsimple <pvr.iptvsimple@noreply.github.com>
Cc: buers <buers@gmx.de>; Mention <mention@noreply.github.com>
Betreff: Re: [kodi-pvr/pvr.iptvsimple] DST Fix (Matrix) (#248)

 

Test builds here:
https://jenkins.kodi.tv/blue/organizations/jenkins/kodi-pvr%2Fpvr.iptvsimple/detail/PR-248/1/artifacts

src/PVRIptvData.cpp
```diff
@@ -804,39 +804,74 @@ int PVRIptvData::GetFileContents(std::string& url, std::string &strContent)
...
+  can be computed(within a constant) by addition of an appropriate constant, integer division
+  by 100, multiplication by 3, and integer division by 4;
+  we can compute correction for any epoch using the same formula we use
+  in the main computation, and can do this with a macro so that this correction is computed
+  at compilation time.
+  In this adaption, epoch is fixed to 1970
+  */
+
+
+static long long my_utctime(int year, int mon, int mday, int hour, int min, int sec)
```
phunkyfish: Why not have this as a private static function on the class? Also, the naming does not follow the same format as every other function (i.e. using Pascal case, e.g. MakeTime() or MakeUTCTime()).

src/PVRIptvData.cpp
```diff
@@ -804,39 +804,74 @@ int PVRIptvData::GetFileContents(std::string& url, std::string &strContent)
...
+  int m, y;
+  y = year + 100;
+  m = mon - 1;
+  if (m < 2)
+  {
+    m += 12;
+    --y;
+  }
+  // compute number of days within constant, assuming appropriate origin
+#define MY_MKTIME(Y,M,D) ((long long)Y*365 + Y/4 - Y/100*3/4 + (M+2)*153/5 + D)
```
phunkyfish: A private inline function on the class would be better assuming the parent function is also class scoped instead of using a macro. Any casts should c++ style casts.

src/PVRIptvData.cpp
```diff
@@ -804,39 +804,74 @@ int PVRIptvData::GetFileContents(std::string& url, std::string &strContent)
...
+  int hour = 0;
+  int min = 0;
+  int sec = 0;
   char sign = '+';
   int hours = 0;
   int minutes = 0;
 
   if (iDateFormat)
-    sscanf(strDate.c_str(), "%04d%02d%02d%02d%02d%02d %c%02d%02d", &timeinfo.tm_year, &timeinfo.tm_mon, &timeinfo.tm_mday, &timeinfo.tm_hour, &timeinfo.tm_min, &timeinfo.tm_sec, &sign, &hours, &minutes);
+    sscanf(strDate.c_str(), "%04d%02d%02d%02d%02d%02d %c%02d%02d",
```
phunkyfish: Not sure this needs to be split over multiple lines, the rest of the code does not do this. But I guess only preference.

src/PVRIptvData.cpp
```diff
@@ -804,39 +804,74 @@ int PVRIptvData::GetFileContents(std::string& url, std::string &strContent)
...
-  std::time(&current_time);
-  long offset = 0;
-#ifndef TARGET_WINDOWS
-  offset = -std::localtime(&current_time)->tm_gmtoff;
-#else
-  _get_timezone(&offset);
-#endif // TARGET_WINDOWS
-
-  long offset_of_date = (hours * 60 * 60) + (minutes * 60);
+    sscanf(strDate.c_str(), "%02d.%02d.%04d%02d:%02d:%02d",
```
phunkyfish: Not sure this needs to be split over multiple lines, the rest of the code does not do this. But I guess only preference.

src/PVRIptvData.cpp
```diff
@@ -804,39 +804,74 @@ int PVRIptvData::GetFileContents(std::string& url, std::string &strContent)
   return strContent.length();
 }
 
+/*
```
phunkyfish: The comments are larger than the code added. Might be better to just credit the author and source. I.e.:

```
/*
  Adapted from https://stackoverflow.com/a/31533119
  Conversion from UTC date to second, signed 64-bit adjustable epoch version.
  Written by François Grieu, 2015-07-21; public domain. 
  */
```

phunkyfish (changes requested):

phunkyfish:
I've used the test driver that was supplied in the issue and was not able to find any issues with this fix. So functionally it's correct.

matthuisman:
@phunkyfish

I'm not a C++ dev so sort of "winged" the latest changes to what looked good to me.
Not sure what to do with that macro, but pretty sure the other changes are correct.

New builds here:
https://jenkins.kodi.tv/blue/organizations/jenkins/kodi-pvr%2Fpvr.iptvsimple/detail/PR-248/2/artifacts

phunkyfish:
Ok, Give me a bit of time today and I’ll give you the static inline function.

matthuisman:
I think this is it?

```
inline long long AGoodName(int year, int month, int day)
{
  return ((long long)year*365 + year/4 - year/100*3/4 + (month+2)*153/5 + day);
}
```

phunkyfish:
Also GetUTCTime should be static and private.

phunkyfish:
Yes, that’s closer to what you need. It can be static and private also. Maybe call the function `MakeTime`. It can go in the header file.

Then you just need a c++ cast. 

On train at present, will help a bit more once I’m properly online.

ksooo:
No need for a member function  at all. Use a simple stand-alone function and wrap it in an unnamed namespace. No need to clutter any class declaration with stuff of this kind.

phunkyfish:
Now I learn something too 😉

So the unnamed namespace containing GetUTCTime declaration and an inline MakeTime in the header file?

phunkyfish:
For the static cast either:

`static_cast<long long>(your-maketime-calculation);`

Or 

`static_cast<int64_t>(your-maketime-calculation);`

Any reason to prefer long long or int64_t or vice versa?

matthuisman:
This is all a bit over my head.

Would the header file meant to look something like this:
```
private:
  static long long GetUTCTime(int year, int mon, int mday, int hour, int min, int sec);

  inline long long MakeTime(int year, int month, int day)
  {
    return ((long long)year*365 + year/4 - year/100*3/4 + (month+2)*153/5 + day);
  }
```
Not inside any class.
Do you need "inline" when it's in the header file?
Oh, do you even need private if it's not in a class?

Then the function in cpp file something like this:
```
/*
  Adapted from https://stackoverflow.com/a/31533119
  Conversion from UTC date to second, signed 64-bit adjustable epoch version.
  Written by François Grieu, 2015-07-21; public domain. 
*/
static long long GetUTCTime(int year, int mon, int mday, int hour, int min, int sec)
{
  int m, y;
  y = year + 100;
  m = mon - 1;

  if (m < 2)
  {
    m += 12;
    --y;
  }

  return (((MakeTime(y, m, mday) - MakeTime((1970 + 99), 12, 1)) * 24 + hour) * 60 + min) * 60 + sec;
}
```

Where does the casting go?

Did some googling and it tells me not to use unnamed namespace in header file?
https://wiki.sei.cmu.edu/confluence/display/cplusplus/DCL59-CPP.+Do+not+define+an+unnamed+namespace+in+a+header+file

phunkyfish:
Ya, I think you’re right. It goes in the cpp file.

Give me 10 mins and will try properly from a computer instead of phone.

ksooo:
no change in the header needed

changes in cpp file:
``` c++

namespace
{
/*
  Adapted from https://stackoverflow.com/a/31533119
  Conversion from UTC date to second, signed 64-bit adjustable epoch version.
  Written by François Grieu, 2015-07-21; public domain. 
*/

long long MakeTime(int year, int month, int day)
{
  return static_cast<long long<(year * 365 + year / 4 - year / 100 * 3 / 4 + (month + 2) * 153 / 5 + day);
}

long long GetUTCTime(int year, int mon, int mday, int hour, int min, int sec)
{
  int m = mon - 1;
  int y = year + 100;

  if (m < 2)
  {
    m += 12;
    --y;
  }

  // compute number of days within constant, assuming appropriate origin
  return (((MakeTime(y, m, mday) - MakeTime(1970 + 99, 12, 1)) * 24 + hour) * 60 + min) * 60 + sec;
}

} // unnamed namespace
```

phunkyfish:
Thanks @ksooo, you learn something new everyday...

ksooo:
`PVRIptvData::ParseDateTime` is returning `int`. So, why the `long long` conversion stuff?

ksooo:
unnamed namespaces is what static non-member functions were "yesterday". Intended to implement non-member stuff that shall be kept private to a cpp file.

phunkyfish:
@matthuisman so you can remove the cast completely. That would have you covered until 2038 ;)

```c++
namespace
{
/*
  Adapted from https://stackoverflow.com/a/31533119
  Conversion from UTC date to second, signed 64-bit adjustable epoch version.
  Written by François Grieu, 2015-07-21; public domain. 
*/

int MakeTime(int year, int month, int day)
{
  return year * 365 + year / 4 - year / 100 * 3 / 4 + (month + 2) * 153 / 5 + day;
}

int GetUTCTime(int year, int mon, int mday, int hour, int min, int sec)
{
  int m = mon - 1;
  int y = year + 100;

  if (m < 2)
  {
    m += 12;
    --y;
  }

  // compute number of days within constant, assuming appropriate origin
  return (((MakeTime(y, m, mday) - MakeTime(1970 + 99, 12, 1)) * 24 + hour) * 60 + min) * 60 + sec;
}

} // unnamed namespace
```

phunkyfish:
@ksooo any need to make MakeTime inline or is that inferred by the unnamed namespace?

ksooo:
I nowadays trust the compilers to know better than I if inlining makes sense.

phunkyfish:
@buers could you run the test driver using the modified code again? Would do it myself but won't be able to till sometime next week. It should be fine but would be good to check it once more.

buers:
Hi,
first of all: I am very grateful to @matthuisman for helping out to get that patch in. (I will try to set things up here, to be able to do myself in the future). And to all comments. 

I had also tested (briefly) binary builds of the patched addon, and they worked as expected.

I am a bit astonished about this discussion. Sorry, if I misunderstand. Seems, I had cluttered namespace by my implementation. But this is not the case at all. The "private helper function" to convert to UTC seconds since epoch start really could only be seen inside one file. No need to put it in any header (which possibly makes it visuable outside) - why? Also no forward declaration of the prototype required (IHMO would make things worse - 2 places to make changes instead of one).

BTW. On purpose I decided to use long long instead of int64_t. Long long is perfectly suitable. I will analyse the code again, but certainly now the GetUTCTime() and the MakeTime() functions again have a year 2038 problem and it really doesn't win anything. Note that this code in long long in my tests was more than an order of magnitude faster than the original (wrong) code. I also decided to not use struct tm. So this small helper function is totally independent of anything environmental. It does not need any type or other help PVRIptvData, so it shouldn't show itself there. Isn't that about as "private" and not cluttering any namespace, as it can get?

I will check, if the now modified code is correct until 2038 later.

The correct fix would be, to make ParseDateTime() return time_t instead of int. This would mean to change at least one header. Without any experience and good knowledge of Kodi source tree, I just wanted to prepare that fix (there is some time left :-). If I have seen enough of the rest of the Kodi source, actually that int will be assigned to time_t later anyway (what else, one could ask).

I agree with the naming changes, thanks. Sorry, that I did not recognize, that you are using camel back nameing conventions. Also getting rid of that "private macro", which really was a hack (but was used correctly), certainly is a good thing. Not sure, if it could mean, that compilers will not find all const expression evaluations at compile time (however, in my fast test with high optimization, everything looks fine - even when not, this really won't slow down the code inside its context).

I had tried to carefully comment / rephrase the existing comment to the function now named GetUTCTime(). This seems to be gone (perhaps partitionally, because it is not true anymore - ie. 2038). Maybe, my phrasing was not optimal or good. But now, important information is missing. (The clever method by François Grieu is some sort of black magic anyway, and without giving some idea, this will look confusing …). I.e. that month starts at 1 not 0 etc. Other comments don't fit anymore after the changes, i.e. // compute number of days within constant, assuming appropriate origin would fit to the MakeTime() function, but not to the return statement.

I don't see how std::chronos could help to make the code cleaner.

All that said. my suggestion
- use the new names and the function instead of that macro
- review comments (but keep some more, than what seems to be there now)
- Keep long long, add a comment that PVRIptvData::ParseDateTime should return time_t, not int, and the the rest of the code is already prepared to do so.
- I will get rid of the old C-Style cast to long Long
- Carefully test it again

Do we really need that namespace stuff?
Did you change any headers?
Can anybody please give the link, where I can see the current code alltogether? (Sorry, when I have missed this).

My first patches showed "clean" diffs here on my local Disc. However, when Matt received them, he had horrible long diffs due to spacing changes. Any idea, how to avoid this? Before creating the patch, I had read coding guidelines, but now, I don't find them anymore fast. https://kodi.wiki/view/Official:Code_guidelines_and_formatting_conventions has a dead link to https://codedocs.xyz/xbmc/xbmc/code_guidelines.html. 

I had adjusted indention to 2 and not to use tabs in the Editor. Or maybe, this was a problem/consequence of using pastebin.com?

BTW. Does anybody have a good pointer, how to compile all this. I started with https://github.com/xbmc/xbmc/blob/master/docs/README.Windows.md
and https://github.com/kodi-pvr/pvr.iptvsimple/blob/Matrix/README.md
But here already the directory conventions used in the two links seem to differ.

ksooo:
(cough)... seeing walls of text... for what exactly?

matthuisman:
@phunkyfish @ksooo @buers
I have updated this pull request with the changes.

I put the code at the top of the file as that made more sense to me than mixing the class code with the unnmaed namespace code?

It seems to have failed to build on IOS platforms though.
grep: cmake/addons/.success: No such file or directory
Not sure if this is issue with the code or something wrong with jenkins?

Latest builds here:
https://jenkins.kodi.tv/blue/organizations/jenkins/kodi-pvr%2Fpvr.iptvsimple/detail/PR-248/4/artifacts/

buers:
Not sure, what you intend to accomplish with your question @ksooo. My post had comments, questions, reasoning, what I think is wrong with the changes, that were done to the fix I had suggested. Well, you are the "member", I have no tag at all. I had some enthuisiasm to fix an old bug - it vanished. I have no intention to waste my free time with discussing such comments.

matthuisman:
Here is a new test code (should work).
https://pastebin.com/WcLhP1NL

**This code tests between the current implementation and the patched implementation**

I've only tested here:
https://www.onlinegdb.com/online_c++_compiler
(which won't have any fails as will be linux and dst bug is windows only)

@buers
You can see the current pull-request (patch) here:
https://github.com/kodi-pvr/pvr.iptvsimple/pull/248/files

And the entire file here:
https://github.com/kodi-pvr/pvr.iptvsimple/blob/150b0229ef3ed98382613f9587236c7f7169b53c/src/PVRIptvData.cpp

Everything you said has basically been done.
There has been 0 changes to the header file.

It appears startTime and endTime are both time_t.
https://github.com/kodi-pvr/pvr.iptvsimple/blob/master/src/PVRIptvData.h#L32

ParseDateTime is only used for these two variables.
https://github.com/kodi-pvr/pvr.iptvsimple/blob/master/src/PVRIptvData.cpp#L265

So, I think we could do

```
time_t iTmpStart = ParseDateTime(strStart);
time_t iTmpEnd = ParseDateTime(strStop);
```

then update ParseDateTime to return time_t

But I think possibly we should do a separate pull-request for that?
And just keep this request for fixing the DST bug.

ksooo:
> So, I think we could do
>
> time_t iTmpStart = ParseDateTime(strStart);
> time_t iTmpEnd = ParseDateTime(strStop);
>
> then update ParseDateTime to return time_t
>

yes, plus change the return type of newly introduced local functions to time_t.

> But I think possibly we should do a separate pull-request for that?
> And just keep this request for fixing the DST bug.

 No separate PR needed. Just change that in this PR.

ksooo:
Guys, to speed things up: Does anybody mind if I open a new PR, based on this PR, containing *all* changes needed (original code adjusted to match our coding rules, return type fix, addon version bump and changelog) to get this in fast? This discussion and this many iterations over the code and what not is far too time consuming... It's a so simple and straight- forward change.... I cannot see why we should spend so much (more) time on this.

matthuisman:
Yes. Please do :)

phunkyfish:
@ksooo I have some time today, if you’d like me to create the new 2 PRs I’m happy to do so.

ksooo:
> I have some time today, if you’d like me to create the new 2 PRs I’m happy to do so.

 I will do that. :-)

 And the backport PR will be created once we merged the matrix PR. I don't see any benefit in editing 2 PRs again and again. ;-)

phunkyfish:
Cool, no worries 😉

ksooo:
Superseded by #255

PR #251, opened 2019-05-14
PVRIptvData: Fix considering DST in timestamp conversion

palinek:
1. Don't use the undocumented tm_gmtoff attribute (implementation specific)
2. Unify the logic for all platforms
3. Consider the diff between UTC and local time at the the particular
timestamp


ksooo:
Superseded by #255

PR #253, opened 2019-05-28
fixed strValue sizeof

matjaz:


ksooo:
Please provide a meaningful commit message and PR description.

ksooo:
What is this supposed to fix? For me, the original code looks right.

src/client.cpp
```diff
@@ -343,17 +343,15 @@ PVR_ERROR GetChannelStreamProperties(const PVR_CHANNEL* channel, PVR_NAMED_VALUE
...
+    PVR_STRCPY(properties[0].strValue, m_currentChannel.strStreamURL.c_str());
     *iPropertiesCount = 1;
     if (!m_currentChannel.properties.empty())
     {
       for (auto& prop : m_currentChannel.properties)
       {
-        strncpy(properties[*iPropertiesCount].strName, prop.first.c_str(),
-                sizeof(properties[*iPropertiesCount].strName) - 1);
-        strncpy(properties[*iPropertiesCount].strValue, prop.second.c_str(),
-                sizeof(properties[*iPropertiesCount].strName) - 1);
```
matjaz, line 356: here it should be `sizeof(properties[*iPropertiesCount].strValue)` instead of strName

ksooo: Good catch. Seems to be a copy / past bug.

PR #255, opened 2019-06-14
Fix wrong EPG times due to DST on Windows. Closes #158.

### LINKED ISSUE #158: [Windows] - Wrong EPG Times due to DST (1 Hour Out)

ksooo:
Supersedes #248, #251

@buers @matthuisman @phunkyfish @palinek good to go?

### DIFF
--- a/pvr.iptvsimple/addon.xml.in
+++ b/pvr.iptvsimple/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.iptvsimple"
-  version="4.0.1"
+  version="4.0.2"
   name="PVR IPTV Simple Client"
   provider-name="nightik">
   <requires>@ADDON_DEPENDS@</requires>
--- a/pvr.iptvsimple/changelog.txt
+++ b/pvr.iptvsimple/changelog.txt
@@ -1,3 +1,6 @@
+v4.0.2
+- Fix wrong EPG times due to DST on Windows
+
 v4.0.1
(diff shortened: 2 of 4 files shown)


ksooo:
Jenkins error on iOS and macOS is unrelated (fallout from https://github.com/xbmc/xbmc/pull/16218).

phunkyfish (approved):
Looks good to me

PR #238, opened 2019-04-22
Remove channels loaded notification

allixx:
resolves #222

PR #259, opened 2019-06-17
Remove channels loaded notification (Leia Backport)

matthuisman:
https://github.com/kodi-pvr/pvr.iptvsimple/pull/238 back-ported to Leia


ksooo:
I only want bugfixes for Leia.

Issue #264, opened 2019-08-18
Crash when loading epg data


skydrome:
Matrix Nightly 20190816 - arm64-v8
addon versions 4.1.0 4.2.0 both crash / **4.0.2 works**

kodi log: https://paste.debian.net/plain/1096352
logcat: https://paste.debian.net/plain/1096353

Crashes after playlist is loaded and epg starts to load
Everything works normally until I add an epg file

Logfile doesnt show anything that stands out so im not sure how else to debug it

I also tested on win64 with version 4.2.0 and it works on windows

phunkyfish:
Do you have the kodi crash log?

skydrome:
android doesnt produce crash dumps, just what logcat CrashDescriptorUtil gives

phunkyfish:
I think I've found the issue referenced above. Update to v4.2.1 once it’s out.

skydrome:
Thanks! That fixed it

PR #265, opened 2019-08-18
Fix nullptr initialisation

phunkyfish:
v4.2.1

- Fix nullptr initialisation


phunkyfish:
Will merge this shortly. Silly mistake while refactoring, reported here: https://github.com/kodi-pvr/pvr.iptvsimple/issues/264

PR #276, opened 2019-09-03
Recompile for 6.1.0 PVR Addon API compatibility

phunkyfish:
v4.4.0
- Update: Recompile for 6.1.0 PVR Addon API compatibility


phunkyfish:
Kodi PR: https://github.com/xbmc/xbmc/pull/16558

PR #281, opened 2019-09-21
Reload interval

phunkyfish:
v4.6.0
- Added: M3U auto refresh
- Update: update github links for Matrix
- Update: Remove dependency on p8-platform util, thread and mutex
- Added: Switch EPGTimeshift setting to hours instead of minutes

### DIFF
--- a/src/PVRIptvData.cpp
+++ b/src/PVRIptvData.cpp
@@ -28,6 +28,8 @@
 #include "iptvsimple/Settings.h"
 #include "iptvsimple/utilities/Logger.h"
 
+#include <chrono>
+
 using namespace ADDON;
 using namespace iptvsimple;
 using namespace iptvsimple::data;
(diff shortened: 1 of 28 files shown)


src/PVRIptvData.h
```diff
@@ -34,6 +31,9 @@
 
 #include <atomic>
 
+#include <kodi/libXBMC_pvr.h>
```
ksooo, line 36: Other addons do not consider "kodi" as system headers. Why this change?

src/PVRIptvData.cpp
```diff
@@ -57,10 +58,27 @@ bool PVRIptvData::Start()
 
 void PVRIptvData::Process()
 {
+  unsigned int refreshTimer = 0;
+  time_t lastRefreshTimeSeconds = time(nullptr);
```
ksooo: std::time

src/PVRIptvData.cpp
```diff
@@ -57,10 +58,27 @@ bool PVRIptvData::Start()
...
+  int lastRefreshHour = Settings::GetInstance().GetM3URefreshHour(); //ignore if we start during same hour
+
   while (m_running)
   {
     std::this_thread::sleep_for(std::chrono::milliseconds(PROCESS_LOOP_WAIT_SECS * 1000));
 
+    time_t currentRefreshTimeSeconds = time(nullptr);
```
ksooo: std:.time

src/PVRIptvData.h
phunkyfish: I found that some kodi headers used system <> and some used regular "".

It made sense to me that as these are not referenced relatively to the local source, i.e. relative to an include path they should be referenced as system headers. Do you think that reasoning is incorrect?

ksooo: I think your reasoning is correct. Maybe you should contact @AlwinEsch and ask him to change that for all official binary addons?

ksooo (approved):

src/PVRIptvData.h
phunkyfish: Will do

PR #269, opened 2019-08-22
Support multiple credits elements, firstAired, year, star rating, episode-num in epg entry, readme update, multiple display-names's and matching channels, plus bugfixes

phunkyfish:
v3.8.0
- Support multiple display-names and case insensitive tvg-id is always first, next tvg-name and then channel name find order
- Support full timeshift range of -12 to +14 hours
- Some providers incorrectly use tvg-ID instead of tvg-id
- Channel groups not being clared when reading next channel
- support episode-num for both xmltv_ns and onscreen systems in epg entry
- Update readme for settings, supported M3U and XMLTV formats and genres
- support star rating in epg entry
- support firstAired and year in epg entry
- Update OSX build scripts
- support multiple actor/director/writers elements in epg entry
- URLEncode and append .png ext for remote logos built from channel name
- Added: Timing for Playlist and EPG load

PR #282, opened 2019-09-22
Speed up EPG loading

vpeter4:
With #269 there are two regex which are evaluated on every function call. In my case loading time was increased from 800msec to 10sec. This PR replaces those regex with static ones.
For Leia branch.

### DIFF
--- a/pvr.iptvsimple/addon.xml.in
+++ b/pvr.iptvsimple/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.iptvsimple"
-  version="3.8.2"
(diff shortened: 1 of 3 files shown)


phunkyfish:
Thanks for this vpeter. Can you please create the PR for Matrix first?

Once that's merged the Leia version can go in.

BTW, you also need to modify the changelog and addon.xml.in. Just increment the patch number as this is a fix.

vpeter4:
Done.

vpeter4:
Aligned with Matrix.

Now loading is so fast that I don't even see notification anymore :-)

phunkyfish (approved):
Looks good. Will merge once CI is happy.

PR #283, opened 2019-09-22
Speed up EPG loading

vpeter4:
With #269 there are two regex which are evaluated on every function call. In my case loading time was increased from 800msec to 10sec. This PR replaces those regex with static ones.

### DIFF
--- a/src/iptvsimple/data/EpgEntry.cpp
+++ b/src/iptvsimple/data/EpgEntry.cpp
@@ -213,7 +213,8 @@ bool EpgEntry::UpdateFrom(rapidxml::xml_node<>* channelNode, const std::string& 
   const std::string dateString = GetNodeValue(channelNode, "date");
   if (!dateString.empty())
   {
-    if (std::regex_match(dateString, std::regex("^[1-9][0-9][0-9][0-9][0-9][1-9][0-9][1-9]")))
+    static const std::regex dateRegex("^[1-9][0-9][0-9][0-9][0-9][1-9][0-9][1-9]");
+    if (std::regex_match(dateString, dateRegex))
       m_firstAired = static_cast<time_t>(ParseDateTime(dateString));
 
     std::sscanf(dateString.c_str(), "%04d", &m_year);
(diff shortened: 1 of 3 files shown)


phunkyfish:
Matrix is slightly different in that you also need to add the changelog entry to the addon.xml.

Just scroll down and you'll see where.

phunkyfish:
There is another regex where you could do the same around line 216 if you don't mind?

vpeter4:
Hope it is ok (building only Leia here).

phunkyfish:
Yes, I can test the matrix PR for you.

vpeter4:
This brings time down from 750ms to 530ms for me :)

src/iptvsimple/data/EpgEntry.cpp
```diff
@@ -213,7 +214,7 @@ bool EpgEntry::UpdateFrom(rapidxml::xml_node<>* channelNode, const std::string&
   const std::string dateString = GetNodeValue(channelNode, "date");
   if (!dateString.empty())
   {
-    if (std::regex_match(dateString, std::regex("^[1-9][0-9][0-9][0-9][0-9][1-9][0-9][1-9]")))
+    if (std::regex_match(dateString, dateRegex))
```
phunkyfish, line 217: Why not move this to the line before it is used?

src/iptvsimple/data/EpgEntry.cpp
```diff
@@ -310,10 +311,12 @@ bool EpgEntry::ParseXmltvNsEpisodeNumberInfo(const std::string& episodeNumberStr
 
 bool EpgEntry::ParseOnScreenEpisodeNumberInfo(const std::string& episodeNumberString)
 {
-  const std::string text = std::regex_replace(episodeNumberString, std::regex("[ \\txX_\\.]"), "");
+  static const std::regex numRegex("[ \\txX_\\.]");
+  static const std::regex epRegex("^[sS]([0-9][0-9]*)[eE][pP]?([0-9][0-9]*)$");
```
phunkyfish: I guess the same here

phunkyfish (changes requested):

phunkyfish:
Nice, half a second is reasonable ;)

vpeter4:
Like this?

phunkyfish:
Yes, looks good.

phunkyfish (approved):
Once everything passes CI I will merge this. AppVeyor can take a while.

phunkyfish:
Actually can you just squash this to a single commit?

vpeter4:
As you wish :)

phunkyfish:
Great, thanks for much for the contribution!

phunkyfish:
If you want to replicate in Leia Pr will review that next.

Issue #295, opened 2019-10-23
Try pugixml instead of rapidxml for C++11 feature support


phunkyfish:
@vpeter4 

Would you be able to test the XML performance if I submit the PR?

vpeter4:
Sure I can. Patiently waiting for it :)

phunkyfish:
It would mean installing a Matrix nightly somewhere to run the tests though.

Is that doable?

vpeter4:
Ah, this could be a problem. From what I see LibreELEC is still on Kodi 18. So I can't build pvr client to test.
Only builds from @MilhouseVH has Kodi 19 but I don't know how to build that version. So I'm out for now. Sorry.

phunkyfish:
Ok, but do you have a dev environment, says Mac or Windows? That would do.

I only have short test files for XMLTV which would be no good for any perf tests.

vpeter4:
If you make me Windows 64bit pvr client I can test. But I don't have build environment for Windows to make it myself.

phunkyfish:
I can do that. Jenkins will build the zip for us as part of the PR.

phunkyfish:
Ok, here you go. You will find the zip for Win64 here: https://jenkins.kodi.tv/blue/organizations/jenkins/kodi-pvr%2Fpvr.iptvsimple/detail/PR-297/10/artifacts

phunkyfish:
Please also check the quality of the data. From some basic tests it appears ok but best to double check on a real dataset.

vpeter4:
old v4.7.0
```
2019-10-25 18:26:15.961 T:3736  NOTICE: AddOnLog: PVR IPTV Simple Client: pvr.iptvsimple - iptvsimple::PlaylistLoader::LoadPlayList Playlist Loaded - 3 (ms)
2019-10-25 18:26:15.961 T:3736  NOTICE: AddOnLog: PVR IPTV Simple Client: pvr.iptvsimple - iptvsimple::PlaylistLoader::LoadPlayList - Loaded 106 channels.
2019-10-25 18:26:17.212 T:6060  NOTICE: AddOnLog: PVR IPTV Simple Client: pvr.iptvsimple - iptvsimple::Epg::LoadChannelEpgs - Loaded '65' EPG channels.
[...]

PR #297, opened 2019-10-24
Move from RapidXML to Pugixml which supports c++11 features + tvg-id fixes

phunkyfish:
v4.7.1
- Update: Switch from rapixml to pugixml (for C++11 syntax)
- Fixed: tvg-id does not match a channel if no display-names in xmltv
- Fixed: Combine multiple XMLTV channels sharing same id

Pugixml is very similar to rapidxml but supports C++11 syntax allowing cleaner code.

https://github.com/kodi-pvr/pvr.iptvsimple/issues/295


phunkyfish:
@2small @Sorien @allixx 

Any chance you can test this from a new feature perspective on Kodi Matrix?

Test builds are here for the standard kodi platforms: https://jenkins.kodi.tv/blue/organizations/jenkins/kodi-pvr%2Fpvr.iptvsimple/detail/PR-297/12/artifacts

Just to make sure nothing that was working is broken after changing the underlying XML engine.

allixx:
tested https://github.com/phunkyfish/pvr.iptvsimple/commit/b241214ef025fab919962b3b065643b2db8dea10 with kodi g96737bf on linux x86_64. builds and runs fine (I have a basic channel list without EPG though).
homedir over nfs probably explains a relatively significant load time.

>2019-10-26 12:09:15.681 T:560  NOTICE: AddOnLog: PVR IPTV Simple Client: pvr.iptvsimple - LoadPlayList Playlist Loaded - 211 (ms)
>2019-10-26 12:09:15.681 T:560  NOTICE: AddOnLog: PVR IPTV Simple Client: pvr.iptvsimple - LoadPlayList - Loaded 477 channels.

phunkyfish:
Thanks for testing @allixx !!!

src/iptvsimple/data/ChannelEpg.cpp
```diff
@@ -34,8 +34,11 @@ bool ChannelEpg::UpdateFrom(const xml_node& channelNode, Channels& channels)
     return false;
 
   bool foundChannel = false;
+  bool haveDisplayNames = false;
   for (const auto& displayNameNode : channelNode.children("display-name"))
   {
+    bool haveDisplayNames = true;
```
ksooo: This doesn't make sense. I guess intention was to assign true to the var declared outside the for loop (line37)?!

ksooo (changes requested):
One bug that needs to be fixed & I hope the new lib was carefully tested to be functionally equivalent to rapidxml.

src/iptvsimple/data/ChannelEpg.cpp
phunkyfish: Oops, so my test would always have passed ;)

Nice find, thanks.

phunkyfish:
I had several users test the latest features/additions with the new lib to make sure everything still worked as expected.

ksooo (approved):

PR #317, opened 2019-11-28
Catchup/Archive Support

phunkyfish:
v4.9.0
- Added: Catchup support (a.k.a Archive)
- Added: Missing check for HLS stream type
- Update: Readme
- Fixed: strings.po comments

### DIFF
--- a/pvr.iptvsimple/resources/language/resource.language.en_gb/strings.po
+++ b/pvr.iptvsimple/resources/language/resource.language.en_gb/strings.po
@@ -259,7 +259,60 @@ msgctxt "#30067"
 msgid "Use FFMpeg http reconnect options if possible"
 msgstr ""
 
-#empty strings from id 30068 to 30599
+#empty strings from id 30068 to 30099
+
+#label-category: catchup
+#label-group: Catchup - Catchup
+msgctxt "#30100"
+msgid "Catchup"
+msgstr ""
+
+#label: Catchup - enableCatchup
+msgctxt "#30101"
+msgid "Enable catchup"
+msgstr ""
+
+#label: Catchup - catchupQueryFormat
+msgctxt "#30102"
+msgid "Query format string"
+msgstr ""
+
+#label: Catchup - catchupDays
+msgctxt "#30103"
+msgid "Catchup window"
+msgstr ""
+
+#label: Catchup - allChannelsSupportCatchup
+msgctxt "#30104"
+msgid "All channels support catchup"
+msgstr ""
+
+#label-group: Catchup - Watch From EPG
+msgctxt "#30105"
+msgid "Watch From EPG"
+msgstr ""
+
+#label: Catchup - catchupPlayEpgAsLive
+msgctxt "#30106"
+msgid "Play from EPG in Live TV mode (using timeshift)"
+msgstr ""
+
+#label: Catchup - catchupWatchEpgBeginBuffer
+msgctxt "#30107"
+msgid "Buffer before program start"
+msgstr ""
+
+#label: Catchup - catchupWatchEpgEndBuffer
+msgctxt "#30108"
+msgid "Buffer after program end"
+msgstr ""
+
+#label: Catchup - catchupOnlyOnFinishedProgrammes
+msgctxt "#30109"
+msgid "Catchup only available on finished programmes"
+msgstr ""
+
+#empty strings from id 30110 to 30599
 
 #############
 # help info #
@@ -445,5 +498,54 @@ msgstr ""
 
 #help: Advanced - useFFmpegReconnect
 msgctxt "#30685"
-msgid "Note this can only apply to http/https streams that are processed by libavformat (e.g. M3u8/HLS). Using libavformat can be specified in an M3U file by setting the property `inputstreamclass` as `inputstream.ffmpeg`. I.e. adding the line: `#KODIPROP:inputstreamclass=inputstream.ffmpeg`. If this opton is not enabled it can still be enabled per stream/channel by adding a kodi property, i.e.: `#KODIPROP:http-reconnect=true`."
+msgid "Note this can only apply to http/https streams that are processed by libavformat (e.g. M3u8/HLS). Using libavformat can be specified in an M3U file by setting the property `inputstreamclass` as `inputstream.ffmpeg` for kodi's internal implementation or `inputstream.ffmpegdirect` for addon version. I.e. adding the line: `#KODIPROP:inputstreamclass=inputstream.ffmpeg`. If this option is not enabled it can still be enabled per stream/channel by adding a kodi property, i.e.: `#KODIPROP:http-reconnect=true`."
+msgstr ""
+
+#empty strings from id 30686 to 30699
+
+#help info - Catchup
+
+#help-category: Catchup
+msgctxt "#30700"
+msgid "Catchup settings for viewing archived live programmes. Allows the option of 'Play programme' when viewing EPG entry info."
+msgstr ""
+
+#help: Catchup - enableCatchup
+msgctxt "#30701"
+msgid "Should be enabled if there are channels supporting catchup in your M3U list."
+msgstr ""
+
+#help: Catchup - catchupQueryFormat
+msgctxt "#30702"
+msgid "A query format string (provider dependent) allowing timestamp information to be appended to a URL to denote when to catchup from. E.g. `&cutv={Y}-{m}-{d}T{H}:{M}:{S}`, which allows year, month, day, hour minute and second to be inserted to give: `&cutv=2019-11-26T22:00:32`. If the M3U entry using has a catchup mode of 'default' or 'append' and a 'catchup-source' tag is provided in the M3U entry this setting will be ignored."
+msgstr ""
+
+#help: Catchup - catchupDays
+msgctxt "#30703"
+msgid "A number of days into the past in which it is possible to catchup on a programme. Can be overidden in an M3U entry using a 'catchup-days' tag."
+msgstr ""
+
+#help: Catchup - allChannelsSupportCatchup
+msgctxt "#30704"
+msgid "If enabled it is assumed that all channels support catchup. If there are no catchup specific tags in the M3U entries then the stream URL will be used as the source, and the 'Query format string' and 'Catchup window' number of days will come from the addon settings. If this option is disabled then a M3U entry must have at least a 'catchup=\"default\"' or 'catchup=\"append\"' tag to enable catchup."
+msgstr ""
+
+#help: Catchup - catchupPlayEpgAsLive
+msgctxt "#30705"
+msgid "When disabled any catchup programme from the past will be played like a video (bounded by start and end times). If enabled, it will instead act like a live stream with timeshift, also allowing the ability to skip back and forward programmes."
+msgstr ""
+
+#help: Catchup - catchupWatchEpgBeginBuffer
+msgctxt "#30706"
+msgid "The amount of buffer to give before the playback start point of an EPG entry that will be watched as a video."
+msgstr ""
+
+#help: Catchup - catchupWatchEpgEndBuffer
+msgctxt "#30707"
+msgid "The amount of buffer to give after the playback end point of an EPG entry that will be watched as a video."
+msgstr ""
+
+#help: Catchup - catchupOnlyOnFinishedProgrammes
+msgctxt "#30708"
+msgid "When selected from the EPG the current live programme cannot be watched as catchup until finished."
 msgstr ""
--- a/pvr.iptvsimple/resources/language/resource.language.en_gb/strings.po
+++ b/pvr.iptvsimple/resources/language/resource.language.en_gb/strings.po
@@ -18,335 +18,432 @@ msgstr ""
 
 #settings labels
 
+#label: General - m3uPathType
+#label: EPG Settings - epgPathType
+#label: Channel Logos - logoPathType
 msgctxt "#30000"
 msgid "Location"
 msgstr ""
 
+#label-option: General - m3uPathType
+#label-option: General - epgPathType
+#label-option: General - logoPathType
+#label-option: General - genresPathType
 msgctxt "#30001"
 msgid "Local path (include local network)"
 msgstr ""
 
+#label-option: General - m3uPathType
+#label-option: General - epgPathType
+#label-option: General - logoPathType
+#label-option: General - genresPathType
 msgctxt "#30002"
 msgid "Remote path (Internet address)"
 msgstr ""
 
+#label-option: General - m3uRefreshMode
 msgctxt "#30003"
 msgid "Disabled"
 msgstr ""
 
+#label-option: General - m3uRefreshMode
 msgctxt "#30004"
 msgid "Repeated refresh"
 msgstr ""
 
+#label-option: General - m3uRefreshMode
 msgctxt "#30005"
 msgid "Once per day"
 msgstr ""
 
 #empty strings from id 30006 to 30009
 
+#label-category: general
+#label-group: General - General
 msgctxt "#30010"
 msgid "General"
 msgstr ""
 
+#label: General - m3uPath
 msgctxt "#30011"
 msgid "M3U playlist path"
 msgstr ""
 
+#label: General - m3uUrl
 msgctxt "#30012"
 msgid "M3U playlist URL"
 msgstr ""
 
+#label: General - startNum
 msgctxt "#30013"
 msgid "Start channel number"
 msgstr ""
 
+#label: General - numberByOrder
 msgctxt "#30014"
 msgid "Only number by channel order in M3U"
 msgstr ""
 
+#label: General - m3uRefreshMode
 msgctxt "#30015"
 msgid "Auto refresh mode"
 msgstr ""
 
+#label: General - m3uRefreshIntervalMins
 msgctxt "#30016"
 msgid "Refresh interval"
 msgstr ""
 
+#label: General - m3uRefreshHour
 msgctxt "#30017"
 msgid "Refresh hour (24h)"
 msgstr ""
 
+#label-group: General - Refresh
 msgctxt "#30018"
 msgid "Refresh"
 msgstr ""
 
+#label-category: epgsettings
 msgctxt "#30019"
 msgid "EPG Settings"
 msgstr ""
 
+#label-group: EPG Settings - EPG Settings
 msgctxt "#30020"
 msgid "EPG"
 msgstr ""
 
+#label: EPG Settings - epgPath
 msgctxt "#30021"
 msgid "XMLTV path"
 msgstr ""
 
+#label: EPG Settings - epgUrl
 msgctxt "#30022"
 msgid "XMLTV URL"
 msgstr ""
 
+#label: EPG Settings - epgTSOverride
 msgctxt "#30023"
 msgid "Apply time shift to all channels"
 msgstr ""
 
+#label: EPG Settings - epgTimeShift
 msgctxt "#30024"
 msgid "EPG time shift"
 msgstr ""
 
+#label: General - m3uCache
 msgctxt "#30025"
 msgid "Cache m3u at local storage"
 msgstr ""
 
+#label: EPG Settings - epgCache
 msgctxt "#30026"
 msgid "Cache XMLTV at local storage"
 msgstr ""
 
+#format-label: EPG Settings - epgTimeShift
 msgctxt "#30027"
 msgid "{0:.1f} hours"
 msgstr ""
 
 #empty strings from id 30028 to 30029
 
+#label-category: channellogos
+#label-group: Channel Logos - Channel Logos
 msgctxt "#30030"
 msgid "Channel Logos"
 msgstr ""
 
+#label: Channel Logos - logoPath
 msgctxt "#30031"
 msgid "Channel logos folder"
 msgstr ""
 
+#label: Channel Logos - logoBaseUrl
 msgctxt "#30032"
 msgid "Channel logos base URL"
 msgstr ""
 
 #empty strings from id 30033 to 30039
 
+#label-group: Channel Logos - XMLTV Logos Options
 msgctxt "#30040"
 msgid "XMLTV Logos Options"
 msgstr ""
 
+#label: Channel Logos - logoFromEpg
 msgctxt "#30041"
 msgid "Channels logos from XMLTV"
 msgstr ""
 
+#label-option: Channel Logos - logoFromEpg
 msgctxt "#30042"
 msgid "Ignore"
 msgstr ""
 
+#label-option: Channel Logos - logoFromEpg
 msgctxt "#30043"
 msgid "Prefer M3U"
 msgstr ""
 
+#label-option: Channel Logos - logoFromEpg
 msgctxt "#30044"
 msgid "Prefer XMLTV"
 msgstr ""
 
 #empty strings from id 30045 to 30049
 
+#label-category: genres
+#label-group: Genres - Genres
 msgctxt "#30050"
 msgid "Genres"
 msgstr ""
 
+#label: Genres - useEpgGenreText
 msgctxt "#30051"
 msgid "Use genre text from XMLTV when mapping genres"
 msgstr ""
 
+#label: Genres - genresPath
 msgctxt "#30052"
 msgid "Genres path"
 msgstr ""
 
+#label: Genres - genresUrl
 msgctxt "#30053"
 msgid "Genres URL"
 msgstr ""
 
 #empty strings from id 30054 to 30059
 
+#label-category: advanced
 msgctxt "#30060"
 msgid "Advanced"
 msgstr ""
 
+#label-group: Advanced - Multicast relay
 msgctxt "#30061"
 msgid "Multicast relay (udpxy)"
 msgstr ""
 
+#label: Advanced - transformMulticastStreamUrls
 msgctxt "#30062"
 msgid "Transform multicast stream URLs"
 msgstr ""
 
+#label: Advanced - udpxyHost
 msgctxt "#30063"
 msgid "Relay hostname or IP address"
 msgstr ""
 
+#label: Advanced - udpxyPort
 msgctxt "#30064"
 msgid "Relay port"
 msgstr ""
 
+#label-group: Advanced - Streaming
 msgctxt "#30065"
 msgid "Streaming"
 msgstr ""
 
+#label: Advanced - useInputstreamAdaptiveforHls
 msgctxt "#30066"
 msgid "Use inputstream.adaptive for m3u8 (HLS) streams"
 msgstr ""
 
+#label: Advanced - useFFmpegReconnect
 msgctxt "#30067"
 msgid "Use FFMpeg http reconnect options if possible"
 msgstr ""
 
 #empty strings from id 30068 to 30599
+
+#############
+# help info #
+#############
+
 #help info - General
 
+#help-category: General
 msgctxt "#30600"
 msgid "General settings required for the addon to function."
 msgstr ""
 
+#help: General - m3uPathType
 msgctxt "#30601"
 msgid "Select where to find the M3U resource. The options are: [Local path] - A path to an M3U file whether it be on the device or the local network; [Remote path] - A URL specifying the location of the M3U file."
 msgstr ""
 
+#help: General - m3uPath
 msgctxt "#30602"
 msgid "If location is [Local path] this setting must contain a valid path for the addon to function."
 msgstr ""
 
+#help: General - m3uUrl
 msgctxt "#30603"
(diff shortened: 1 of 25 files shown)


Sorien:
is it ready for testing?

phunkyfish:
No, not in the slightest, I'm afraid. Will be another while yet. Assuming the straight port works then it needs to be integrated with the other addon features. This is step one of many.

phunkyfish:
@Sorien very early days but if you want to try it and see what works? Doc's are written for finished product whereas first version has no timeshift whatsoever (only start and stop).

If you use Play EPG as live TV setting the playback should be correct but it will play as if the UI is on live TV. Looks weird but allows you to access the channel and guide menus from the Live TV.

Builds:
https://jenkins.kodi.tv/blue/organizations/jenkins/kodi-pvr%2Fpvr.iptvsimple/detail/PR-317/12/artifacts

Sorien:
i tried it with my addon and it's working pretty well

phunkyfish:
Did you use it with "Play as live" enabled/disabled or both?

Sorien:
i'm not sure what you mean 

I'm creating playlist like 
```
#EXTINF:-1 tvg-id="2458735488" tvg-logo="jednotkahd.png" catchup-days="7" catchup-source="plugin://plugin.video.sl/?stationid=2458735488&askpin=False&utc=${start}",Jednotka HD
plugin://plugin.video.sl/?id=rzxqQ-kzUkG3x2PGEaxnFAAAAAE&askpin=False
```
and then i can play it with "play programme" button from epg

phunkyfish:
So it's playing as a video? Or with a similar OSD to live TV?

There is an option in the addon settings for this which you can enable/disable.

Should also switch to this build, made a few bug fixes: https://jenkins.kodi.tv/blue/organizations/jenkins/kodi-pvr%2Fpvr.iptvsimple/detail/PR-317/14/artifacts

Sorien:
its playing like video stream (it's dash stream protected by widevine so i'm using  inputstreamadaptive for playback) 

when i try to use "Play from Epg In Live Mode" from iptvsimple settings stream wont start 
```
CVideoPlayer::OpenInputStream - error opening [pvr://guide/0028/2019-12-02 18:00:00.epg]
```

phunkyfish:
Can you maybe try something from several hours ago? And maybe the new build?

However I have not tested this with adaptive as I don't have a compatible provider so it may just be buggy. Could you check the logged EPG URL's and see if the generated ones are the same for both options?

Sorien:
i can play multiple days back without problem but it will not find correct program from time to time it seems that like your 5min substraction makes small problem

phunkyfish:
You can set that to zero in settings. But I'm not sure why it was cause an issue unless your provider only allows catchup on programme boundaries. 

But please verify the URLs generated in the logs to see if they are correct. If you find bugs I'm happy to try and fix them.

Sorien:
after more testing i found that almost none of iptv providers in my country supports playing programme that is currently live in tv only finished ones. Could we have some settings to mark current programme as not playable like `catchup-only-finished=True|False`

phunkyfish:
You mean as part of the M3U entry or in the addon settings?

Sorien:
I'd say that it would be better to have it as a part of m3u.

phunkyfish:
Ok, if I ever get this working for real I will add this.

phunkyfish:
@Sorien I added an addon setting to only allow catchup on the programmes that have finished. Can you try it?

Sorien:
@phunkyfish thx, I'll try it in a few days.

pvr.iptvsimple/resources/language/resource.language.en_gb/strings.po
```diff
@@ -350,3 +408,52 @@ msgstr ""
 msgctxt "#30685"
 msgid "Note this can only apply to http/https streams that are processed by libavformat (e.g. M3u8/HLS). Using libavformat can be specified in an M3U file by setting the property `inputstreamclass` as `inputstream.ffmpeg`. I.e. adding the line: `#KODIPROP:inputstreamclass=inputstream.ffmpeg`. If this opton is not enabled it can still be enabled per stream/channel by adding a kodi property, i.e.: `#KODIPROP:http-reconnect=true`."
```
ksooo: > Using libavformat can be specified in an M3U file by setting the property `inputstreamclass` as `inputstream.ffmpeg`. I.e. adding the line: `#KODIPROP:inputstreamclass=inputstream.ffmpeg`. 

inputstream.ffmpegdirect ?

> If this opton is not 

typo. option

pvr.iptvsimple/resources/language/resource.language.en_gb/strings.po
```diff
@@ -350,3 +408,52 @@ msgstr ""
...
+msgstr ""
+
+#help: Catchup - catchupQueryFormat
+msgctxt "#30702"
+msgid "A query format string (provider dependent) allowing timestamp information to be appended to a URL to denote when to catchup from. E.g. `&cutv={Y}-{m}-{d}T{H}:{M}:{S}`, which allows year, month, day, hour minute and second to be inserted to give: `&cutv=2019-11-26T22:00:32`. If the M3U entry using has a catchup mode of 'default' or 'append' and a 'catchup-source' tag is provided in the M3U entry this setting will be ignored."
+msgstr ""
+
+#help: Catchup - catchupDays
+msgctxt "#30703"
+msgid "A number of days into the past in which it is possible to catchup on a show. Can be overidden in an M3U entry using a 'catchup-days' tag."
```
ksooo: > catchup on a show

 I thought right word in en_gb would be "programme", no?

pvr.iptvsimple/resources/language/resource.language.en_gb/strings.po
```diff
@@ -350,3 +408,52 @@ msgstr ""
...
+msgstr ""
+
+#help: Catchup - allChannelsSupportCatchup
+msgctxt "#30704"
+msgid "If enabled it is assumed that all channels support catchup. If there are no catchup specific tags in the M3U entries then the stream URL will be used as the source, and the 'Query format string' and 'Catchup window' number of days will come from the addon settings. If this option is disabled then a M3U entry must have at least a 'catchup=\"default\"' or 'catchup=\"append\"' tag to enable catchup."
+msgstr ""
+
+#help: Catchup - catchupPlayEpgAsLive
+msgctxt "#30705"
+msgid "When disabled any catchup show from the past will be played like a video (bounded by start and end times). If enabled, it will instead act like a live stream with timeshift, also allowing the ability to skip back and forward programmes."
```
ksooo: show vs. programme

pvr.iptvsimple/resources/language/resource.language.en_gb/strings.po
```diff
@@ -350,3 +408,52 @@ msgstr ""
...
+msgstr ""
+
+#help: Catchup - catchupWatchEpgEndBuffer
+msgctxt "#30707"
+msgid "The amount of buffer to give after the playback end point of an EPG entry that will be watched as a video."
+msgstr ""
+
+#help: Catchup - catchupOnlyOnFinishedProgrammes
+msgctxt "#30708"
+msgid "When selected from the EPG the current live show cannot be watched as catchup until finished."
```
ksooo: show vs. programme

src/client.cpp
```diff
@@ -131,6 +134,7 @@ ADDON_STATUS ADDON_Create(void* hdl, void* props)
     m_currentStatus = ADDON_STATUS_LOST_CONNECTION;
     return m_currentStatus;
   }
+  m_catchupController = &m_data->GetCatchupController();
```
ksooo, line 137: Is it guaranteed that the raw pointer stored in this member will be valid as long as you access it using the member?

src/client.cpp
```diff
@@ -240,57 +244,33 @@ PVR_ERROR GetChannels(ADDON_HANDLE handle, bool bRadio)
...
-        StreamUtils::SetStreamProperty(properties, iPropertiesCount, propertiesMax, prop.first, prop.second);
-    }
+    m_catchupController->ResetCatchupState(); // TODO: we need this currently until we have a way to know the stream stops.
+
+    // We always call the catchup controller regardless so it can cleanup state
+    // whether or not it supports catchup in case there is any houskeeping to do
+    std::map<std::string, std::string> catchupProperties;
+    m_catchupController->ProcessChannelForPlayback(m_currentChannel, catchupProperties);
+
+    std::string catchupUrl = m_catchupController->GetCatchupUrl(m_currentChannel);
```
ksooo: `const std::string` ? Or is it modified by `StreamUtils::SetAllStreamProperties`?

src/client.cpp
```diff
@@ -324,12 +304,70 @@ PVR_ERROR GetChannelGroupMembers(ADDON_HANDLE handle, const PVR_CHANNEL_GROUP& g
...
 }
 
+PVR_ERROR IsEPGTagPlayable(const EPG_TAG* tag, bool* bIsPlayable)
+{
+  if (!m_data)
+    return PVR_ERROR_SERVER_ERROR;
+  if (!settings.IsCatchupEnabled())
+    return PVR_ERROR_NOT_IMPLEMENTED;
+
+  const time_t now = time(nullptr);
```
ksooo: std::time ?

src/client.cpp
```diff
@@ -324,12 +304,70 @@ PVR_ERROR GetChannelGroupMembers(ADDON_HANDLE handle, const PVR_CHANNEL_GROUP& g
...
+    *iPropertiesCount = 0;
+    Logger::Log(LEVEL_DEBUG, "%s - GetPlayEpgAsLive is %s", __FUNCTION__, settings.CatchupPlayEpgAsLive() ? "enabled" : "disabled");
+
+    std::map<std::string, std::string> catchupProperties;
+    if (settings.CatchupPlayEpgAsLive())
+      m_catchupController->ProcessEPGTagForTimeshiftedPlayback(*tag, m_currentChannel, catchupProperties);
+    else
+      m_catchupController->ProcessEPGTagForVideoPlayback(*tag, m_currentChannel, catchupProperties);
+
+    std::string catchupUrl = m_catchupController->GetCatchupUrl(m_currentChannel);
```
ksooo: `const std::string`? Or is it modified by `StreamUtils::SetAllStreamProperties`?

src/iptvsimple/CatchupController.cpp
```diff
@@ -0,0 +1,436 @@
...
+
+  if (m_controlsLiveStream)
+  {
+    if (m_resetCatchupState)
+    {
+      m_resetCatchupState = false;
+      if (channel.IsCatchupSupported())
+      {
+        m_timeshiftBufferOffset = Settings::GetInstance().GetCatchupDaysInSeconds(); //offset from now to start of catchup window
+        m_timeshiftBufferStartTime = time(nullptr) - Settings::GetInstance().GetCatchupDaysInSeconds(); // now - the window size
```
ksooo: std::time ?

src/iptvsimple/CatchupController.cpp
```diff
@@ -0,0 +1,436 @@
...
+    m_timeshiftBufferStartTime = 0;
+    m_timeshiftBufferOffset = 0;
+
+    m_fromEpgTag = true;
+  }
+}
+
+void CatchupController::ProcessEPGTagForVideoPlayback(const EPG_TAG& epgTag, Channel& channel, std::map<std::string, std::string>& catchupProperties) // TODO: should be a const channel, create a StreamManager instead
+{
+  TestAndStoreStreamType(channel); // TODO: StreamManager
```
ksooo, line 143: I do spot so many new TODO's in this PR. I hope the implementation is actually usable? ;-)

src/iptvsimple/CatchupController.cpp
```diff
@@ -0,0 +1,436 @@
...
+
+  if (m_catchupStartTime > 0)
+    m_playbackIsVideo = true;
+}
+
+void CatchupController::SetCatchupInputStreamProperties(bool playbackAsLive, const Channel& channel, std::map<std::string, std::string>& catchupProperties)
+{
+  catchupProperties.insert({"epgplaybackaslive", playbackAsLive ? "true" : "false"});
+
+  catchupProperties.insert({"inputstream.ffmpegdirect.is_realtime_stream", "true"});
```
ksooo, line 188: How does the addon behave if inputstream.ffmpegdirect is not installed? Would be cool if the user would at least get a meaningful error message with instructions how to "fix" this.

src/iptvsimple/CatchupController.cpp
```diff
@@ -0,0 +1,436 @@
...
+  Logger::Log(LEVEL_DEBUG, "catchup_buffer_start_time - %s", std::to_string(m_catchupStartTime).c_str());
+  Logger::Log(LEVEL_DEBUG, "catchup_buffer_end_time - %s", std::to_string(m_catchupEndTime).c_str());
+  Logger::Log(LEVEL_DEBUG, "catchup_buffer_offset - %s", std::to_string(m_timeshiftBufferOffset).c_str());
+  Logger::Log(LEVEL_DEBUG, "timezone_shift - %s", std::to_string(channel.GetTvgShift()).c_str());
+}
+
+void CatchupController::TestAndStoreStreamType(Channel& channel)
+{
+  // We need to find out what type of stream this is, so let's construct a catchup URL to test with
+  std::string streamTestUrl = GetStreamTestUrl(channel);
```
ksooo: const std::string?

src/iptvsimple/CatchupController.cpp
```diff
@@ -0,0 +1,436 @@
...
+//   m_resetCatchupState = true;
+//   Logger::Log(LEVEL_NOTICE, "CloseLiveStream - true");
+// }
+
+namespace
+{
+
+void FormatOffset(time_t tTime, std::string &urlFormatString)
+{
+  const std::string regexStr = ".*(\\{offset:(\\d+)\\}).*";
```
ksooo: "const std::string directly in this unnamed namespace" to not construct it on every call?

src/iptvsimple/CatchupController.h
```diff
@@ -0,0 +1,87 @@
...
+    // Current programme details
+    time_t m_programmeStartTime;
+    time_t m_programmeEndTime;
+    std::string m_programmeTitle;
+    unsigned int m_programmeUniqueChannelId = 0;
+    int m_programmeChannelTvgShift = 0;
+
+    bool m_controlsLiveStream = false;
+    iptvsimple::Epg& m_epg;
+    std::mutex* m_mutex;
```
ksooo: `= nullptr`?

src/iptvsimple/CatchupController.h
```diff
@@ -0,0 +1,87 @@
...
+    time_t m_catchupStartTime = 0;
+    time_t m_catchupEndTime = 0;
+    time_t m_timeshiftBufferStartTime = 0;
+    long long m_timeshiftBufferOffset = 0;
+    bool m_resetCatchupState = false;
+    bool m_playbackIsVideo = false;
+    bool m_fromEpgTag = false;
+
+    // Current programme details
+    time_t m_programmeStartTime;
```
ksooo: = 0 ?

src/iptvsimple/CatchupController.h
```diff
@@ -0,0 +1,87 @@
...
+    time_t m_programmeEndTime;
```
ksooo: = 0 ?

src/iptvsimple/data/EpgEntry.cpp
```diff
@@ -184,8 +184,8 @@ bool EpgEntry::UpdateFrom(const xml_node& channelNode, const std::string& id, in
   if (!GetAttributeValue(channelNode, "start", strStart) || !GetAttributeValue(channelNode, "stop", strStop))
     return false;
 
-  long long tmpStart = ParseDateTime(strStart);
-  long long tmpEnd = ParseDateTime(strStop);
+  long long tmpStart = ParseDateTime(strStart) + (86400 * 66); //2020-01-30
```
ksooo: 2020-01-30 ?

ksooo (commented):
Awesome.

pvr.iptvsimple/resources/language/resource.language.en_gb/strings.po
phunkyfish: Either actually. One is kodi's internal impl the other the addon.

phunkyfish: I've updated the text to reflect that.

pvr.iptvsimple/resources/language/resource.language.en_gb/strings.po
phunkyfish: Yup

src/client.cpp
phunkyfish: Yes, everywhere it is used has a check for m_data which is only deleted on addon destroy.

src/client.cpp
phunkyfish: Should be const, you're right.

src/iptvsimple/CatchupController.cpp
phunkyfish: Many TODO's but still functional!!!!!

src/iptvsimple/CatchupController.cpp
phunkyfish: Hopefully to be fixed by Alwin by allowing a function to check if an addon is installed.

If the addon is not installed it will go through the inputstream factory and be handled by kodi internally. Thing is the user won't know (bar the missing video player controls) until we get that function.

pvr.iptvsimple/resources/language/resource.language.en_gb/strings.po
ksooo: 👍

src/client.cpp
ksooo: Cool. Thought that you had checked that. Just wanted to be sure.

src/iptvsimple/CatchupController.cpp
ksooo: 👍

src/iptvsimple/CatchupController.cpp
phunkyfish: Ya, so really this should be static regex as regex's are expensive to create.

So if `static const std::regex rx(".*(\\{offset:(\\d+)\\}).*");` the same as `const std::regex rx(".*(\\{offset:(\\d+)\\}).*");` in an unnamed namespace?

src/iptvsimple/data/EpgEntry.cpp
phunkyfish: Ha, that's my EPG data hack, should never have been in there!!!

src/iptvsimple/CatchupController.cpp
phunkyfish: I went with const without the static.

ksooo: If you want to prevent that the regex is created each time the function is called -- and this should be the goal here -- you either need to make it static inside the function or to put it outside the function in an unnamed namespace.

phunkyfish: Ah so like:

```
const std::regex rx(".*(\\{offset:(\\d+)\\}).*");

void FormatOffset(time_t tTime, std::string &urlFormatString)
{
  std::cmatch mr;
  if (std::regex_match(urlFormatString.c_str(), mr, rx) && mr.length() >= 3)
  {
    std::string offsetExp = mr[1].first;
    std::string second = mr[1].second;
```

phunkyfish: Done!

phunkyfish:
Should be ready now.

src/iptvsimple/CatchupController.cpp
ksooo: 👍

ksooo: I would give the constant a more speaking (because it is visible inside the whole unnamend namespace now) and all upper-case name (because this is common practice for constant's name).

ksooo: On second thought, static inside the function is imo better, because you don't need the value outside that function. unnamed namespace const would be the choice if you need the value even from outside the function.

phunkyfish: Ok, done, should be good now.

phunkyfish:
Ok, ready to go.

phunkyfish:
Forgot to capitalise it Now it's really ready!

ksooo (approved):
Shove it in.

Sorien:
i have updated iptvsimple from artifacts and kodi (today's nightly win 10 64bit) and when i try to play programme from epg it just crashes kodi https://gist.github.com/Sorien/7073b32b511607bbc8eac9a7bc34f9b3, not sure if its related to your changes couse i can't even play live channels :/

phunkyfish:
Good timing, was just about to hit merge!

Do you have a crashlog too?

Also are the last 2 lines of your log showing the correct URL?

Sorien:
there is no crash log generated i can create memory dump if you want 
give me few mins i'll recheck few things

Sorien:
I corrected my plugin urls and noticed that play programme and switch button seems to crash at same line (should iptvsimple process catchup url when you just switching channel?)

play programme from epg
https://gist.github.com/Sorien/bc5493f6f8bd5286ffb4eead971ae114

memory dump (with procdump)
[13:31:18] Unhandled: C00000FD.STACK_OVERFLOW
[kodi.exe_200202_133118.zip](https://github.com/kodi-pvr/pvr.iptvsimple/files/4144158/kodi.exe_200202_133118.zip)

kodi seems not to fully crash, process is still running but there is no visible ui

switch programme from the channel list
https://gist.github.com/Sorien/e97713a9590ed8dee075f628a91d0750

generated m3u playlist
https://gist.github.com/Sorien/cf5b33f60a2947c12099367d3493887f

phunkyfish:
Do you know how to run a windows debugger?

I can get you a Pdb file.

Sorien:
i hope so

phunkyfish:
PDB: https://mirrors.kodi.tv/test-builds/windows/win64/KodiSetup-20200129-6b070a5e-rtp-source-ip-fix-x64.pdb
EXE: https://mirrors.kodi.tv/test-builds/windows/win64/KodiSetup-20200129-6b070a5e-rtp-source-ip-fix-x64.exe

If you can figure out how to run it this should automatically pickup the addon and tell you when exactly the error happens.

Sorien:
for some reason this version seems to not crash and reveals some bugs in my plugin 

2 things i noticed so far
1. when i try to play livetv channel iptvsimple will use catchup url
2. ${end} its replaced with wrong timestamp ${end} is lower ${start}

DEBUG: XFILE::CPluginDirectory::StartScript - calling plugin Magio Go('plugin://plugin.iptv.magiogo/play-programme-catchup/4146-158064**3300**-158064**0300**','6','','resume:false')

phunkyfish:
I can't manage to repro this on another OS. Can you get the call stack leading up to that call? On osx the equivalent would be a `thread backtrace`.

Also, do you have Catchup URL format string you are using? I can try and figure out why you get end less than start.

EDIT: Apologies, just saw your M3U file so I have the string.

Sorien:
catchup url looks like plugin://plugin.iptv.magiogo/play-programme-catchup/my_programe_id-${start}-${end}

whole m3u is few posts upper ... i corrected my plugin now $end can be lower then $start, now i have working playback from epg and live tv

but to play stream from epg i have to download epg data from web api and process it so switching channel takes like 10sec it's ok for playing catchup but not for live stream

phunkyfish:
> but to play stream from epg i have to download epg data from web api and process it so switching channel takes like 10sec it's ok for playing catchup but not for live stream

Can you explain this one further? Is this with your plugin or an issue with IPTV Simple?

Do you still have a problem with it using the catchup URL for live even if you have catchup only for finished shows enabled?

phunkyfish:
Ok, minor change so it should no longer use the Catchup URL if it's live in your case @Sorien 

Just make sure you enable: Catchup only for finished programmes

Artifacts ready to go.

Sorien:
> Can you explain this one further? Is this with your plugin or an issue with IPTV Simple?

for catchup stream url i have to know programme unique id (url looks like http://my-provider/api/?programme_id=123456789) and so i have to route it through my plugin, plugin connects to http api of my iptv provider to get epg data, parse it and find programme id to construct playable url and this process takes some time ~10s so my programme switching time goes from <1s to ~10s

thx for update now live steam working like expected

Sorien:
when it comes to $(end) $end according docs should be in utc
xmltv:
`<programme channel="4222" start="20200202 10 50 00" stop="20200202 12 55 00">` utc

play from epg in live tv mode - disabled
buffer before program start 5min
buffer after program start 15min

$start should be 02/02/2020 @ 10:45am (UTC)
$end should be 02/02/2020 @ 1:10pm (UTC)

my plugin gets 

$start 1580640300 02/02/2020 @ 10:45am (UTC) ok 
$end 1580632800 02/02/2020 @ 8:40am (UTC) (looks like -15min should be +15) and -4hours???

src/iptvsimple/CatchupController.cpp
```diff
@@ -0,0 +1,429 @@
...
+  FormatTime('Y', dateTime, fomrattedUrl);
+  FormatTime('m', dateTime, fomrattedUrl);
+  FormatTime('d', dateTime, fomrattedUrl);
+  FormatTime('H', dateTime, fomrattedUrl);
+  FormatTime('M', dateTime, fomrattedUrl);
+  FormatTime('S', dateTime, fomrattedUrl);
+  FormatUtc("{utc}", dateTimeEpg, fomrattedUrl);
+  FormatUtc("${start}", dateTimeEpg, fomrattedUrl);
+  FormatUtc("{utcend}", dateTimeEpg + duration, fomrattedUrl);
+  FormatUtc("${end}", dateTimeEpg + duration, fomrattedUrl);
```
Sorien, line 332: it look like duration has negative value

src/iptvsimple/CatchupController.cpp
```diff
@@ -0,0 +1,429 @@
...
+  if (m_catchupStartTime > 0)
+    return BuildUrlFormatString(channel);
+
+  return "";
+}
+
+std::string CatchupController::GetCatchupUrl(const Channel& channel) const
+{
+  if (m_catchupStartTime > 0)
+    return BuildEpgTagUrl(m_catchupStartTime, static_cast<time_t>(m_programmeStartTime - m_programmeEndTime), channel, m_timeshiftBufferOffset);
```
Sorien: m_programmeEndTime - m_programmeStartTime

phunkyfish: Whoops, nice find. My test provider doesn’t use duration, will post an update shortly.

phunkyfish:
> for catchup stream url i have to know programme unique id (url looks like http://my-provider/api/?programme_id=123456789) and so i have to route it through my plugin, plugin connects to http api of my iptv provider to get epg data, parse it and find programme id to construct playable url and this process takes some time ~10s so my programme switching time goes from <1s to ~10s

Is this programme_id part of the xml file that IPTV Simple uses? I’m wondering is there a way this can be passed as part of the URL.

Sorien:
> Is this programme_id part of the xml file that IPTV Simple uses? I’m wondering is there a way this can be passed as part of the URL.

these programe_ids are mostly UUIDs so there is no way how to generate them from iptvsimple side but we could extend xmltv with some new tag that could be added as a replacement to catchup url or add directly catchup url to xmltv

Sorien (approved):

phunkyfish:
> > Is this programme_id part of the xml file that IPTV Simple uses? I’m wondering is there a way this can be passed as part of the URL.
> 
> these programe_ids are mostly UUIDs so there is no way how to generate them from iptvsimple side but we could extend xmltv with some new tag that could be added as a replacement to catchup url or add directly catchup url to xmltv

This is what I was thinking. If they are part of the XMLTV that IPTV Simple loads then maybe that can save that time. The notion of a provider having its own programme ID is quite common. 

So a question. If you have to set the programme_id how does that work with start and end buffer? Because that could mean both the start and end of the video are actually part of a different programme_id.

Sorien:
> So a question. If you have to set the programme_id how does that work with start and end buffer? 
Because that could mean both the start and end of the video are actually part of a different programme_id.

yes they are part of different programme_id

Every programme has its own stream. It's something more like recordings in Kodi, live stream is recorded and cutter to shorter parts (programmes) according to epg start and end timestamps, every part gets own programme_id and you can access that stream by programme_id, there is no continuous playback from one stream to another and you can seek only within that one stream, so it doesn't matter if start/end buffer is 0, 5min, 15min...

phunkyfish:
So start and end buffer should really be set to zero for your use case?

If you want me to add support for an external programme ID in the XMLTV just open a GitHub issue for it. Then you could just have another placeholder in the catchup URL like ${external-id} or something.

PR #323, opened 2020-01-25
User-Agent choice

Be4stElectrjc:
Hi, using the headers passing option found an easy solution for the problem. Simply added on the fly the User-Agent during the stream parsing. From the addon settings you can choose the User-Agent you like. I didn't update all the strings yet for the new UI feature...
Keep in mind that User-Agent change it's only for the channels included in the playlist, so Kodi keeps it's own identity. I am open to discussion, let me know.

### DIFF
--- a/src/PVRIptvData.cpp
+++ b/src/PVRIptvData.cpp
@@ -802,7 +802,21 @@ bool PVRIptvData::LoadPlayList(void)
       channel.iTvgShift         = tmpChannel.iTvgShift;
       channel.bRadio            = tmpChannel.bRadio;
       channel.properties        = tmpChannel.properties;
-      channel.strStreamURL      = strLine;
+      
+      auto propPair = channel.properties.find("http-user-agent");
+      if (propPair != channel.properties.end())
+      {
+        channel.strStreamURL = AddHeaderToStreamUrl(strLine, "user-agent", propPair->second);
+      }
+      else if (!g_userAgent.empty())
+      {
+        channel.strStreamURL = AddHeaderToStreamUrl(strLine, "user-agent", g_userAgent);
+      }
+      else
+      {
+        channel.strStreamURL = strLine;
+      }
+      
       channel.iEncryptionSystem = 0;
 
       iChannelNum++;
@@ -1481,4 +1495,24 @@ int PVRIptvData::GetChannelId(const char * strChannelName, const char * strStrea
     iId = ((iId << 5) + iId) + c; /* iId * 33 + c */
 
   return abs(iId);
+}
+
+std::string PVRIptvData::AddHeaderToStreamUrl (const std::string& url, const std::string& header, const std::string& value) const
+{
+  char separator = '|';
+  size_t found = url.find('|');
+
+  if (found != std::string::npos)
+  {
+    if (url.find(header + '=', found + 1) != std::string::npos)
+    {
+      return url;
+    }
+    else
+    {
+      separator = '&';
+    }
+  }
+
+  return url + separator + header + '=' + value;
 }
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ General settings required for the addon to function.
 * **M3U Play List URL**: If location is `Remote path` this setting must contain a valid URL for the addon to function.
 * **Cache M3U at local storage**: If location is `Remote path` select whether or not the the M3U file should be cached locally.
 * **Numbering channel starts at**: The number to start numbering channels from.
+* **User-Agent**: Select which User-Agent to use if there is not one supplied as a property or as part of the channel stream URL.
 
 ### EPG Settings
 Settings related to the EPG.
--- a/pvr.iptvsimple/addon.xml.in
+++ b/pvr.iptvsimple/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.iptvsimple"
-  version="3.8.8"
+  version="3.9.8"
   name="PVR IPTV Simple Client"
   provider-name="nightik">
   <requires>@ADDON_DEPENDS@</requires>
--- a/pvr.iptvsimple/changelog.txt
+++ b/pvr.iptvsimple/changelog.txt
@@ -1,3 +1,6 @@
+v3.9.8
+- Added: User-Agent support from advanced addon setting
+
 v3.8.8
 - Fixed: tvg-ID not accepted as an M3U tag
 
--- a/pvr.iptvsimple/resources/language/resource.language.en_gb/strings.po
+++ b/pvr.iptvsimple/resources/language/resource.language.en_gb/strings.po
@@ -113,3 +113,7 @@ msgstr ""
 msgctxt "#30044"
 msgid "Prefer XMLTV"
 msgstr ""
+
+msgctxt "#30045"
+msgid "Change User-Agent"
+msgstr ""
--- a/pvr.iptvsimple/resources/settings.xml
+++ b/pvr.iptvsimple/resources/settings.xml
@@ -8,6 +8,7 @@
     <setting id="m3uUrl" type="text" label="30012" default="" visible="eq(-2,1)"/>
     <setting id="m3uCache" type="bool" label="30025" default="true" visible="eq(-3,1)"/>
     <setting id="startNum" type="number" label="30013" default="1" />
+    <setting id="userAgent" type="text" label="30045" />
   </category>
 
   <!-- EPG -->
--- a/src/PVRIptvData.h
+++ b/src/PVRIptvData.h
@@ -121,13 +121,14 @@ protected:
   PVRIptvEpgChannel* FindEpgForChannel(const std::string& id);
   virtual PVRIptvEpgChannel*   FindEpgForChannel(PVRIptvChannel &channel);
   virtual bool                 FindEpgGenre(const std::string& strGenre, int& iType, int& iSubType);
-  virtual bool                 GzipInflate( const std::string &compressedBytes, std::string &uncompressedBytes);
+  virtual bool                 GzipInflate(const std::string &compressedBytes, std::string &uncompressedBytes);
   virtual int                  GetCachedFileContents(const std::string &strCachedName, const std::string &strFilePath,
                                                      std::string &strContent, const bool bUseCache = false);
   virtual void                 ApplyChannelsLogos();
   virtual void                 ApplyChannelsLogosFromEPG();
   virtual std::string          ReadMarkerValue(std::string &strLine, const char * strMarkerName);
   virtual int                  GetChannelId(const char * strChannelName, const char * strStreamUrl);
+  virtual std::string          AddHeaderToStreamUrl(const std::string& url, const std::string& header, const std::string& value) const;
 
 protected:
   virtual void *Process(void);
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -57,6 +57,7 @@ CHelper_libXBMC_pvr   *PVR  = NULL;
 std::string g_strTvgPath    = "";
 std::string g_strM3UPath    = "";
 std::string g_strLogoPath   = "";
+std::string g_userAgent     = "";
 int         g_logoPathType  = 0;
 int         g_iEPGTimeShift = 0;
 int         g_iStartNumber  = 1;
@@ -165,6 +166,10 @@ void ADDON_ReadSettings(void)
   if (XBMC->GetSetting(iPathType ? "logoBaseUrl" : "logoPath", &buffer))
   {
     g_strLogoPath = buffer;
+  }
+  if (XBMC->GetSetting("userAgent", &buffer))
+  {
+    g_userAgent = buffer;
   }
 
   // Logos from EPG
--- a/src/client.h
+++ b/src/client.h
@@ -44,6 +44,7 @@ extern CHelper_libXBMC_pvr          *PVR;
 extern std::string g_strM3UPath;
 extern std::string g_strTvgPath;
 extern std::string g_strLogoPath;
+extern std::string g_userAgent;
 extern int         g_logoPathType;
 extern int         g_iEPGTimeShift;
 extern int         g_iStartNumber;


phunkyfish:
If you want to add this feature you first need to add it to the Matrix version of the addon. Bear in mind that the Matrix version is quite different architecturally. 

In addition the only strings.po file that should be modified en_gb. The others are generated by an automatic system.

ksooo:
And please use more speaking commit messages. Describe the actual change, not which file you modifies.

Be4stElectrjc:
Ok, will do that.

Be4stElectrjc:
Reformatted all the messages etc. 
Do i need to open another PR for Matrix?

phunkyfish:
Yes, you will need a PR for Matrix, bear in mind the code base for the Matrix version is quite different. But we can review this PR first if you would like. Then you will have a better sense what would be required for Matrix.

pvr.iptvsimple/resources/language/resource.language.it_it/strings.po
```diff
@@ -78,7 +78,7 @@ msgstr "Loghi canali"
 
 msgctxt "#30031"
 msgid "Channels Logos Folder"
-msgstr "Cartella  loghi canali"
+msgstr "Cartella loghi canali"
```
phunkyfish: Please remove this change. We only every update the en_gb strings.po file. All other translation files are done by transifix.

phunkyfish (changes requested):
Ok, the changes look ok. You could move the header modification code to it's own function if you like, just a suggestion.

One thing though. The feature will override or add a user-agent protocol option to all channels. Is that really what most users will need? Really there should always be a way to prefer the user-agent in the m3u if there is one. What do you think? That might mean you need more than one setting.

pvr.iptvsimple/resources/language/resource.language.it_it/strings.po
Be4stElectrjc: Resolved.

Be4stElectrjc:
Well, added the option to override the User-Agent in the playlist assuming the user wants total control on the addon. But i can remove it, less code. In this case, the feature will be used if no User-Agent is provided as header in the stream. Implemented the change pretty hardcoded, will move the code in a function to be used for headers handling, like in Matrix.

phunkyfish:
So what’s important is that the behaviour is the same when a user migrates from Leia to Matrix.

As there are already ways to override user-agent in Matrix you should figure out how you want the feature to work there.

I’m only giving you guidance, but always think of the user experience when adding features.

phunkyfish:
Ok, looks good. Let me know once you have the PR ready for Matrix. Once that's in we can look to merge this one.

This will also need a minor version increment in addon.xml and a changelog entry. No rush on this, just needs to happen before merge.

Be4stElectrjc:
Have PR ready for Matrix.

Made modifications to Leia, no hardcoding and letting priority to the playlist like happens in Matrix.

phunkyfish:
Please open a new one. Just put matrix in The title ;)

src/PVRIptvData.cpp
```diff
@@ -1482,3 +1487,26 @@ int PVRIptvData::GetChannelId(const char * strChannelName, const char * strStrea
...
+
+std::string& PVRIptvData::AddHeaderToStreamUrl(std::string& strLine, const std::string& header)
+{
+  size_t found = strLine.find('|');
+
+  if (found != std::string::npos)
+  {
+    size_t present = strLine.find(header + '=', found + 1);
+
+    if (present == std::string::npos)
```
phunkyfish: Did you forget to add the header here?

src/PVRIptvData.cpp
```diff
@@ -1482,3 +1487,26 @@ int PVRIptvData::GetChannelId(const char * strChannelName, const char * strStrea
...
+    {
+      strLine += '&';
+    }
+
+    return strLine;
```
phunkyfish: Why directly modify the reference and then return it?

src/PVRIptvData.cpp
```diff
@@ -1482,3 +1487,26 @@ int PVRIptvData::GetChannelId(const char * strChannelName, const char * strStrea
...
+  }
+  else
+  {
+    strLine += '|';
+  }
+
+  return strLine += header + '=' + g_userAgent;
```
phunkyfish: This statement could just be in the else above as it will only be executed if the else is.

phunkyfish (changes requested):

src/PVRIptvData.cpp
```diff
@@ -802,7 +802,12 @@ bool PVRIptvData::LoadPlayList(void)
       channel.iTvgShift         = tmpChannel.iTvgShift;
       channel.bRadio            = tmpChannel.bRadio;
       channel.properties        = tmpChannel.properties;
-      channel.strStreamURL      = strLine;
+      
+      if (!g_userAgent.empty() && channel.properties.find("http-user-agent") == channel.properties.end())
```
phunkyfish: What about also populating user-agent from the property if it's present?

phunkyfish (changes requested):

src/PVRIptvData.cpp
Be4stElectrjc: Changed the variable to present, considering have header as argument of the function. The header it's added at the end.

src/PVRIptvData.cpp
Be4stElectrjc: Because forgot putting the else.

src/PVRIptvData.cpp
Be4stElectrjc: This statement must be executed in two cases, so instead of writing it twice, i decided to write it once at the end.

src/PVRIptvData.cpp
Be4stElectrjc: It's not needed because all properties from the stream are already taken here.

Be4stElectrjc:
> This will also need a minor version increment in addon.xml and a changelog entry. No rush on this, just needs to happen before merge.

All links in the addon.xml are dead btw, both here and in Matrix...

They should be correct for en_gb which are the only ones that matter.

src/PVRIptvData.cpp
phunkyfish: Yes, but nothing is done with the user-agent property. If just stays in the channel properties and is never used as far as I can see.

src/PVRIptvData.cpp
phunkyfish: Makes sense, although modifying the strLine variable and then returning the result doesn't make sense. I think it will confuse other devs. Maybe just return the value.

src/PVRIptvData.cpp
Be4stElectrjc: It must be used, if not the properties mechanism wouldn't work at all. A thing that can be tested quickly with a playlist containing properties (didn't try).

src/PVRIptvData.cpp
```diff
@@ -1482,3 +1487,28 @@ int PVRIptvData::GetChannelId(const char * strChannelName, const char * strStrea
 
   return abs(iId);
 }
+
+std::string& PVRIptvData::AddHeaderToStreamUrl(std::string& strLine, const std::string& header)
```
phunkyfish: Why not pass the strLine as a const url and just capture the separator char before the return?

So with a signature of:
`std::string PVRIptvData::AddHeaderToStreamUrl(const std::string& url, const std::string& header)`

src/PVRIptvData.cpp
```diff
@@ -1482,3 +1487,28 @@ int PVRIptvData::GetChannelId(const char * strChannelName, const char * strStrea
...
+{
+  size_t found = strLine.find('|');
+
+  if (found != std::string::npos)
+  {
+    size_t present = strLine.find(header + '=', found + 1);
```
phunkyfish: Why not put the find call in the if statement as you don't use the `present` variable other than in the if.

src/PVRIptvData.cpp
```diff
@@ -1482,3 +1487,28 @@ int PVRIptvData::GetChannelId(const char * strChannelName, const char * strStrea
...
+
+    if (present == std::string::npos)
+    {
```
phunkyfish, line 1508: You don't need either set of brackets in this if/else.

src/PVRIptvData.cpp
phunkyfish: And then a `char separator = '|';` which get's used in the return.

src/PVRIptvData.cpp
Be4stElectrjc: That flow never happens. The return it's inside an else indicating the header was already present, so the playlist has priority over user setting. Before in the else block there was the code to erase the header from the current stream line, so it was more clear. Can add comments to the code if needed, not a prob.

src/PVRIptvData.cpp
Be4stElectrjc: Keep in mind the char separator depends about there are or not other headers in the stream line.
For the signature i would keep strLine to be more clear, even if not strictly necessary (opinion). Used return by reference to optimize speed, but for a built-in type it's negligible.

src/PVRIptvData.cpp
Be4stElectrjc: I can save one variable and a line using it directly inside the if statement, for sure. The purpose it's to make the lines shorter.

src/PVRIptvData.cpp
Be4stElectrjc: I need the return only to be executed when the if condition it's not verified. Not in both cases. However, i can spare some brackets for the inner "single line" if/else statements.

src/PVRIptvData.cpp
phunkyfish: But why does this make sense:
`return strLine += header + '=' + g_userAgent;`

Why not just just return `strLine + header + '=' + g_userAgent;`

src/PVRIptvData.cpp
phunkyfish: All the properties mechanism does is make a key value pair available to the backend. It does not necessarily do anything with it. Certainly not a vlc only property.

src/PVRIptvData.cpp
phunkyfish: Ok, but then why return anything at all? Just modify strLine OR return a value. Don’t do both.

Also your function is called add header to stream url but it’s argument is called strLine. Variable names should be relevant to the local scope, not the parent scope.

src/PVRIptvData.cpp
Be4stElectrjc: It's ok for the purpose. Since it's a key-value pair already available from the stream i can't delete or edit a property from the playlist, since has higher priority. The only thing i can do it's to use the setting when it's not available.

phunkyfish:
Here is what I mean about the signature, return type and use a separator char instead. Now the function could be used for adding any header.

```
std::string PVRIptvData::AddHeaderToStreamUrl(const std::string& url, const std::string& header, const std::string& value)
{
  size_t found = strLine.find('|');

  char separator = '|';
  if (found != std::string::npos)
  {
    if (strLine.find(header + '=', found + 1) != std::string::npos)
      return url;
    else
      separator = '&';
  }

  return url + separator + header + '=' + value;
}
```

src/PVRIptvData.cpp
phunkyfish: But then why not this:

```
auto propPair = channel.properties.find("http-user-agent");
if (propPair != channel.properties.end())
  channel.strStreamURL = AddHeaderToStreamUrl(strLine, "user-agent", propPair->second);
else (!g_userAgent.empty())
  channel.strStreamURL = AddHeaderToStreamUrl(strLine, "user-agent", g_userAgent);
```

Then you replicate the same functionality there is in Matrix.

src/PVRIptvData.cpp
Be4stElectrjc: It's not a prob, used this one because it was committed by you on the Matrix branch.

src/PVRIptvData.cpp
Be4stElectrjc: Yeah, meanwhile i could add that unused property of the playlist.

Be4stElectrjc:
I tried to remain consistent with your function in Matrix and be explicit, but you saved my time optimizing it. Nice work.

`const std::string& PVRIptvData::AddHeaderToStreamUrl(const std::string& url, const std::string& header, const std::string& value) const`

The const in the return because we return a const, the const for the function as caution.
Changes on the go.

src/PVRIptvData.cpp
Be4stElectrjc: Yeah, just saw it. 
You're right, it's better with that function name.

src/PVRIptvData.cpp
phunkyfish: I think that is the unused property of the playlist 😉

phunkyfish:
Ok, and then you just need addon.xml and changelog similar to Matrix. I think on Leia the changelog entry is only in the changelog.

src/PVRIptvData.cpp
```diff
@@ -1482,3 +1492,23 @@ int PVRIptvData::GetChannelId(const char * strChannelName, const char * strStrea
 
   return abs(iId);
 }
+
+const std::string& PVRIptvData::AddHeaderToStreamUrl (const std::string& url, const std::string& header, const std::string& value) const
```
phunkyfish: The return type should be `std:string` and not `std::string&` as the returned variable will be on the stack. 

The compiler will optimise so the object is only created once anyway and it will act like a returned reference.

src/PVRIptvData.cpp
```diff
@@ -802,7 +802,17 @@ bool PVRIptvData::LoadPlayList(void)
       channel.iTvgShift         = tmpChannel.iTvgShift;
       channel.bRadio            = tmpChannel.bRadio;
       channel.properties        = tmpChannel.properties;
-      channel.strStreamURL      = strLine;
```
phunkyfish, line 805: You need to keep this line. Otherwise you will never set the stream url if you are not setting user agent.

phunkyfish (changes requested):
Just the bug above to fix and then I think this is good.

src/PVRIptvData.cpp
Be4stElectrjc: `else {channel.strStreamURL = strLine;}`

phunkyfish: That will work too.

src/PVRIptvData.cpp
```diff
@@ -802,7 +802,21 @@ bool PVRIptvData::LoadPlayList(void)
...
+      auto propPair = channel.properties.find("http-user-agent");
+      if (propPair != channel.properties.end())
+      {
+        channel.strStreamURL = AddHeaderToStreamUrl(strLine, "user-agent", propPair->second);
+      }
+      else if (!g_userAgent.empty())
+      {
+        channel.strStreamURL = AddHeaderToStreamUrl(strLine, "user-agent", g_userAgent);
+      }
+	  else
```
phunkyfish: Indentation is wrong. Should be inline with rest of statement.

Be4stElectrjc: Now it's correct.

phunkyfish (approved):
Nice. Well done for putting up with my reviews 😉

Thanks again for contributing!

phunkyfish:
Oops, we actually got the version number wrong. It should have been 3.9.0!

No harm done though 😉

Be4stElectrjc:
You can always correct it, ty for the help.

phunkyfish:
I think it’s ok as it is. I don’t think there will be another Leia release before Matrix.

Thanks again!

PR #327, opened 2020-02-16
New catchup modes

phunkyfish:
v4.11.0
- Added: Support new catchup providers, shift, xc and fs including TS stream support
- Added: Support for timestamp catchup format specifier
- Added: Optional catchup-id per programme from XMLTV
- Added: Add catchup M3U8 examples to README
- Fixed: Live catchup streams that don't support timeshift
- Fixed: Don't build test catchup url for non catchup streams
- Added: Support for offset catchup format specifier
- Update: Add p8-platform to addon depends instead of from kodi

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,7 @@ The format specifiers are substitution based and work as follows:
 - `{M}`: The minute (00-59) of the start date\time.
 - `{S}`: The second (00-59) of the start date\time.
 - `{offset:X}`: The current offset (now - start time) divided by X seconds. Allows conversion to minutes and other time units. The minimum divider is 1, it must be an integer (not 1.5 or 2.25 etc.) and it must be a positive value. E.g. If you need an offset of 720 for a start time of 2 hours ago (2 hours is 7200 seconds), it means your divider is 10: `{offset:10}`. If you need minutes for the same offset you could use: `{offset:60}` which would result in a value of 120.
+- `{catchup-id}`: A programme specific identifier required in the catchup URL, value loaded from XMLTV programme entries. 
 
 Here’s some examples of how the different formats would look:
 - `?utc={utc}&lutc={lutc}`
@@ -253,7 +254,7 @@ General information on the XMLTV format can be found [here](http://wiki.xmltv.or
 
 **Programme elements**
 ```
-  <programme start="20080715003000 -0600" stop="20080715010000 -0600" channel="channel-x">
+  <programme start="20080715003000 -0600" stop="20080715010000 -0600" channel="channel-x" catchup-id="34534590">
     <title>My Show</title>
     <desc>Description of My Show</desc>
     <category>Drama</category>
@@ -273,7 +274,7 @@ General information on the XMLTV format can be found [here](http://wiki.xmltv.or
     <icon src="http://path-to-icons/my-show.png"/>
   </programme>
 ```
-The `programme` element supports the attributes `start`/`stop` in the format `YYYmmddHHMMSS +/-HHMM` and the attribute `channel` which needs to match the `channel` element's attribute `id`.
+The `programme` element supports the attributes `start`/`stop` in the format `YYYmmddHHMMSS +/-HHMM` and the attribute `channel` which needs to match the `channel` element's attribute `id`. There is an extra attribute which is not part of the XMLTV specification called `catchup-id`. Some providers require a programme specific id, this value can be used as part of the [catchup query format string specifiers](#catchup-format-specifiers).
 
 - `title`: The title of the prgramme.
 - `desc`: A descption of the programme.
--- a/README.md
+++ b/README.md
@@ -188,12 +188,15 @@ http://path-to-stream/live/channel-a.ts
 http://path-to-stream/live/channel-b.ts
 #EXTINF:0 catchup="append" catchup-source="&cutv={Y}-{m}-{d}T{H}:{M}:{S}" catchup-days="3",Channel C
 http://path-to-stream/live/channel-c.ts
+#EXTINF:0 tvg-id="channel-d" tvg-name="Channel-D" catchup="default" catchup-days="1" catchup-source="http://yoururl/channeld/video-{utc}-{duration}.m3u8",Channel D
+http://yoururl/channeld/video.m3u8
 ```
 
 *Explanation for Catchup entries*
 - For `Channel A` the stream URL will be used and the Query format string from the addon settings will be appended to construct the  catchup URL.
 - For `Channel B` the stream URL will not be used, instead using catchup-source as the catchup URL.
 - For `Channel C` the stream URL will be used and catchup-source will be appended to form the catchup URL.
+- For `Channel D` this is example of a flussonic style entry.
 
 Note: The minimum required for a channel/stream is an `#EXTINF` line with a channel name and the `URL` line. E.g. a minimal version of the exmaple file above would be:
 
--- a/README.md
+++ b/README.md
@@ -148,6 +148,7 @@ The format specifiers are substitution based and work as follows:
 - `{utc}`: The start time of the programme in UTC format.
 - `${start}`: Same as `{utc}`.
 - `{lutc}`: Current time in UTC format.
+- `${timestamp}`: Same as `{lutc}`.
 - `{utcend}`: The start time of the programme in UTC format + `${duration}`.
 - `${end}`: Same as `{utcend}`.
 - `{duration}`: The programme duration + any start and end buffer (if set).
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ The format specifiers are substitution based and work as follows:
 - `{lutc}`: Current time in UTC format.
 - `{utcend}`: The start time of the programme in UTC format + `${duration}`.
 - `${end}`: Same as `{utcend}`.
-- `${duration}`: The programme duration + any start and end buffer (if set).
+- `{duration}`: The programme duration + any start and end buffer (if set).
 - `{Y}`: The 4-digit year (YYYY) of the start date\time.
 - `{m}`: The month (01-12) of the start date\time.
 - `{d}`: The day (01-31) of the start date\time.
@@ -181,7 +181,18 @@ http://path-to-stream/live/channel-x-hd.ts
 http://path-to-stream/live/channel-y.ts
 #EXTINF:0,Channel Z
 http://path-to-stream/live/channel-z.ts
+#EXTINF:0 catchup="default",Channel A
+http://path-to-stream/live/channel-a.ts
+#EXTINF:0 catchup="default" catchup-source="http://path-to-stream/live/catchup-b.ts&cutv={Y}-{m}-{d}T{H}:{M}:{S}" catchup-days="3",Channel B
+http://path-to-stream/live/channel-b.ts
+#EXTINF:0 catchup="append" catchup-source="&cutv={Y}-{m}-{d}T{H}:{M}:{S}" catchup-days="3",Channel C
+http://path-to-stream/live/channel-c.ts
 ```
+
+*Explanation for Catchup entries*
+- For `Channel A` the stream URL will be used and the Query format string from the addon settings will be appended to construct the  catchup URL.
+- For `Channel B` the stream URL will not be used, instead using catchup-source as the catchup URL.
+- For `Channel C` the stream URL will be used and catchup-source will be appended to form the catchup URL.
 
 Note: The minimum required for a channel/stream is an `#EXTINF` line with a channel name and the `URL` line. E.g. a minimal version of the exmaple file above would be:
 
--- a/README.md
+++ b/README.md
@@ -124,7 +124,13 @@ Addon settings for catchup:
 * **Query format string**: A format string (provider dependent) allowing timestamp information to be appended to a URL to denote when to catchup from. E.g. `&cutv={Y}-{m}-{d}T{H}:{M}:{S}`, which allows year, month, day, hour minute and second to be inserted to give: `&cutv=2019-11-26T22:00:32`. If the M3U entry using has a catchup mode of `default` or `append` and a `catchup-source` tag is provided in the M3U entry this setting will be ignored.
 * **Catchup window**: The number of days into the past in which it is possible to catchup on a show. Can be overidden in an M3U entry using a 'catchup-days' tag.
 * **All channels support catchup**: If enabled it is assumed that all channels support catchup. If there are no catchup specific tags in the M3U entries then the stream URL will be used as the source, and the URL format and catchup days will come from the addon settings. If this option is disabled then a M3U entry must have at least a `catchup="true"` or `catchup="default"` tag to enable catchup.
-* **All channels support catchup**: If enabled it is assumed that all channels support catchup. If there are no catchup specific tags in the M3U entries then the stream URL will be used as the source, and the `Query format string` and `Catchup window` number of days will come from the addon settings. If this option is disabled then a M3U entry must have at least a `catchup="default"` or `catchup="append"` tag to enable catchup.
+* **All channels support catchup using mode**: If enabled it is assumed that all channels support catchup using the selected mode if they do not have catchup tags. In this case the 'Query format string' and 'Catchup window' number of days will come from the addon settings if needed. If this option is disabled then an M3U entry must have at least a `catchup=` tag to enable catchup. The options for how to build the catch URL are:
+    - `Disabled` - Do not assume all channel support catchup.
+    - `Default` - Use catchup source as the full catchup URL, if there is no catchup source use Append mode.
+    - `Append` - Append the catchup source to the channel URL, if there is no catchup source append the `Query format string` instead.
+    - `Shift (SIPTV)` - Append the standard SIPTV catchup string to the channel URL.
+    - `Flussonic` - Build a flussonic URL from the channel URL.
+    - `Xtream codes` - Build an Xtream codes URL from the channel URL.
 * **Play from EPG in Live TV mode (using timeshift)**: When disabled any catchup show from the past will be played like a video (bounded by start and end times). If enabled, it will instead act like a live stream with timeshift, also allowing the ability to skip back and forward programmes.
 * **Buffer before programme start**: The amount of buffer to give before the playback start point of an EPG entry that will be watched as a video.
 * **Buffer after programme end**: The amount of buffer to give after the playback end point of an EPG entry that will be watched as a video.
@@ -151,15 +157,16 @@ The format specifiers are substitution based and work as follows:
 - `${timestamp}`: Same as `{lutc}`.
 - `{utcend}`: The start time of the programme in UTC format + `${duration}`.
 - `${end}`: Same as `{utcend}`.
-- `{duration}`: The programme duration + any start and end buffer (if set).
 - `{Y}`: The 4-digit year (YYYY) of the start date\time.
 - `{m}`: The month (01-12) of the start date\time.
 - `{d}`: The day (01-31) of the start date\time.
 - `{H}`: The hour (00-23) of the start date\time.
 - `{M}`: The minute (00-59) of the start date\time.
 - `{S}`: The second (00-59) of the start date\time.
+- `{duration}`: The programme duration + any start and end buffer (if set).
+- `{duration:X}`: The programme duration (as above) divided by X seconds. Allows conversion to minutes and other time units. The minimum divider is 1, it must be an integer (not 1.5 or 2.25 etc.) and it must be a positive value. E.g. If you have a duration of 7200 seconds and you need 2 hours (2 hours is 7200 seconds), it means your divider is 3600: `{offset:3600}`. If you need minutes for the same duration you could use: `{offset:60}` which would result in a value of 120.
 - `{offset:X}`: The current offset (now - start time) divided by X seconds. Allows conversion to minutes and other time units. The minimum divider is 1, it must be an integer (not 1.5 or 2.25 etc.) and it must be a positive value. E.g. If you need an offset of 720 for a start time of 2 hours ago (2 hours is 7200 seconds), it means your divider is 10: `{offset:10}`. If you need minutes for the same offset you could use: `{offset:60}` which would result in a value of 120.
-- `{catchup-id}`: A programme specific identifier required in the catchup URL, value loaded from XMLTV programme entries. 
+- `{catchup-id}`: A programme specific identifier required in the catchup URL, value loaded from XMLTV programme entries.
 
 Here’s some examples of how the different formats would look:
 - `?utc={utc}&lutc={lutc}`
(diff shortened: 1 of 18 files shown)


SerhiyRomanov:
So, the results of my tests for Flussonic format
**System:**
```
Starting Kodi (19.0-ALPHA1 Git:20200216-ae7c3d1216). Platform: Linux x86 64-bit
Using Release Kodi x64 build
Kodi compiled 2020-02-16 by GCC 7.4.0 for Linux x86 64-bit version 4.15.18 (266002)
Running on Ubuntu 18.04.4 LTS, kernel: Linux x86 64-bit version 4.15.0-76-generic
FFmpeg version/source: 4.0.4-Kodi
```
Also was installed inputstream.ffmpegdirect addon from your last PR.

**M3U8 playlist:**
#EXTM3U
#EXTINF:0 catchup="default" catchup-days="1" catchup-source="http://ch01.spr24.net/151/archive-{utc}-{duration}.m3u8?token=my_token",1+1
http://ch01.spr24.net/151/index.m3u8?token=my_token

Catchup works correctly. Timeshift also works in both cases (I mean timeshift for Live TV and timeshift for the archived programme)

**M3U playlist**
#EXTM3U 
#EXTINF:0 tvg-name="1+1" catchup="default" catchup-days="1" catchup-source="http://ch01.spr24.net/151/timeshift_abs-{utc}.ts?token=my_token",1+1
http://ch01.spr24.net/151/mpegts?token=my_token

Catchup works correctly. Timeshift doesn't even available. Is it okay?

So, Flussonic format works correctly, if set up "catchup-source" in the playlist for each item. Looking forward to the ability to set this format for whole playlist from addon's settings.

phunkyfish:
What’s happens when you use a `catchup=“flussonic”` tag instead of default?

Note that `catchup-source` in the M3U will be ignored for this catchup mode.

SerhiyRomanov:
**M3U8 playlist:**
```
#EXTM3U
#EXTINF:0 catchup="flussonic" catchup-days="3",1+1
http://ch01.spr24.net/151/index.m3u8?token=my_token
```
Catchup works correctly. Timeshift also works in both cases

**M3U playlist**
```
#EXTM3U
#EXTINF:0 catchup="flussonic" catchup-days="3",1+1
http://ch01.spr24.net/151/mpegts?token=my_token
```
Catchup doesn't even available (i don't see button "Play programme").  Timeshift also doesn't even available. What's wrong with my M3U playlist?

phunkyfish:
For the TS based catchup stream I think that is because only m3u8 is enabled for timeshift currently.

If you switch on debug mode it should log the catchup url. Does that look correct? Just search for flussonic in the log.

Please post the debug log.

Once we are happy with the new options and catchup modes I can look at enabling timeshift for TS streams.

EDIT: I think I know why the TS stream did not work even for catchup. Will update the PR later.

Sorien:
@phunkyfish thx for catchup-id, it looks really nice :) ... I hope to find some time to test it in the near future

phunkyfish:
> @phunkyfish thx for catchup-id, it looks really nice :) ... I hope to find some time to test it in the near future

No problem. The recent changes probably mean a good retest of everything is needed ;)

phunkyfish:
@SerhiyRomanov if you want to try the latest builds: https://jenkins.kodi.tv/blue/organizations/jenkins/kodi-pvr%2Fpvr.iptvsimple/detail/PR-327/6/artifacts/

These should resolve the issues you had. Note that you also require the latest ffmpegdirect build: https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.ffmpegdirect/detail/PR-14/9/artifacts

SerhiyRomanov:
Thank you @phunkyfish! Unfortunately, I use Ubuntu for testing so I'm gonna build new versions tomorrow and will let you know after that.

phunkyfish:
No worries 😉

In the meantime I’ll look at how it’s possible to enable timeshift of TS streams.

phunkyfish:
Let me know if the additional commit works for TS streams.

SerhiyRomanov:
@phunkyfish 
HLS catchup and timeshift appears broken 
`NOTICE: AddOnLog: Inputstream FFmpeg Direct: inputstream.ffmpegdirect property: inputstream.ffmpegdirect.catchup_url_format_string = http://ch01.spr24.net/151/index.m3u8-${start}-{duration}.m3u8?token=my_token`
instead of correct http://ch01.spr24.net/151/archive-1350274200-4200.m3u8?token=my_token

When I select "All channels support catchup using mode" = "Flussonic" in settings then Catchup not available for both type of list. So it only works if I set it manually via M3U-tags

I would like to provide you with my playlists for testing, I suppose it will speed up the development process. What do you think about it?

Mr-Groch:
Looks like problem is with index.**m3u8**-${start}-{duration}.m3u8:

`http://ch01.spr24.net/151/index.m3u8-${start}-{duration}.m3u8?token=my_token`

It should be:

`http://ch01.spr24.net/151/index-${start}-{duration}.m3u8?token=my_token`

Using archive list type is risky for setups with mono/video playlists type. I do not recommend using hardcoded archive list type. It should be the same type as in channel url - if this is index type, then index type in catchup url means the same as archive

phunkyfish:
Can you give the original channel url?

All I really need is the original and translated url to fix this. It’s just a minor issue with the regex.

SerhiyRomanov:
@phunkyfish 
Here is parts of log where my URL are present:

```
NOTICE: AddOnLog: PVR IPTV Simple Client: pvr.iptvsimple - GetChannelStreamProperties - Live Stream URL: http://ch01.spr24.net/151/index.m3u8?token=mytoken

NOTICE: AddOnLog: Inputstream FFmpeg Direct: inputstream.ffmpegdirect property: inputstream.ffmpegdirect.catchup_url_format_string = http://ch01.spr24.net/151/index.m3u8-${start}-{duration}.m3u8?token=mytoken
```

phunkyfish:
@SerhiyRomanov should be fixed now

SerhiyRomanov:
@phunkyfish Yes, Catchup/Timeshift with M3U8 are fixed now.

phunkyfish:
Addon setting should be fixed now too.

phunkyfish:
TS Streams fixed now too.

Last issue is following redirects. Currently the iptvsimple addon does not do that when using ffmpegdirect. That may take a while to fix.

Mr-Groch:
I can confirm - catchup from EPG and timeshift from live works with XC TS list

Mr-Groch:
Unfortunately I have found an issue with timeshifting from live. When I timeshift backwards 1 minute, after minute of watching stream will stop

phunkyfish:
That might be because duration is bounded by `now`.

I would guess that flussonic TS won’t have this problem as no duration is used.

SerhiyRomanov:
@phunkyfish Yes, flussonic TS works correctly and doesn't stop.

Mr-Groch:
Would be possible to add reconnect functionality to timeshift? For each reconnect, start time would be old end time and end time would be current `now`.

phunkyfish:
> Unfortunately I have found an issue with timeshifting from live. When I timeshift backwards 1 minute, after minute of watching stream will stop

And if you go back 5 mins does it play for 5 and then stop?

phunkyfish:
> Would be possible to add reconnect functionality to timeshift? For each reconnect, start time would be old end time and end time would be current `now`.

The addon doesn’t currently get informed when the stream stops (there is no API for it). So very hard to do currently.

Mr-Groch:
> And if you go back 5 mins does it play for 5 and then stop?

Yes

phunkyfish:
> > And if you go back 5 mins does it play for 5 and then stop?
> 
> Yes

Ok, then this should be reasonable to fix. What is a sensible default duration, say 4 hours?

Will all the APIs work with this?

Mr-Groch:
This will not work. I have made PHP online converters for XC and Flussonic lists for Archive Client addon to inject catchup-source to lists. I have hardcoded 4 hours duration to URL - unfortunately both backends were cutting end time to `now`...

phunkyfish:
So this is an inherent fault with the backends.

Is there a different api that can be called when we are close to `now`. Maybe something offset based. So it doesn’t treat it like an archive, instead like live with an offset.

Mr-Groch:
For **Flussonic** there is special URL for rewinding, but I haven't tested it... Don't know if this will serve continuously or just to `now` - time from request... 

`http://flussonic:8080/channel/index-1350274200-now.m3u8`
(timestamp like in catchup url, `index` as usual can be list_type from live url, `now` is const string)

Also found something like this:
`http://flussonic:8080/channel/index-timeshift_abs-1350274200.m3u8`
totally not documented, but I think meaning of each field is as usual

Don't know anything similar for XC...

SerhiyRomanov:
@Mr-Groch @phunkyfish Here is documentation https://flussonic.com/doc/digital-video-recording-dvr/recording-broadcasts-catchup-tv
Let's read it together)

Mr-Groch:
Yes, know this link, unfortunately not all possibilities are clearly documented (like information that different list types can be used in URLs), but this is good start point for searching for information

phunkyfish:
We are already using timeshift_abs for flussonic TS streams. I can add a commit to try it for M3U8.

phunkyfish:
Ok, try the latest commit and see how it works for m3u8.

SerhiyRomanov:
@phunkyfish Seems like problem with regex
```
NOTICE: AddOnLog: PVR IPTV Simple Client: pvr.iptvsimple - GetEPGTagStreamProperties - EPG Catchup URL: http://ch01.spr24.net/151/index-timeshift_abs-1582269000.m3u8?token=my_token
NOTICE: AddOnLog: Inputstream FFmpeg Direct: inputstream.ffmpegdirect property: inputstream.ffmpegdirect.catchup_url_format_string = http://ch01.spr24.net/151/index-timeshift_abs-${start}.m3u8?token=my_token
```
But the correct format is http://flussonic-ip/111/timeshift_abs-1429829884.m3u8
**index-** is redundant

//By the way, http://ch01.spr24.net/151/timeshift_abs-1582269000.m3u8?token=my_token works for Flussonic M3U8 with my IPTV provider :)

Mr-Groch:
Yes I can confirm that. I have also pure flussonic list now available. If in channel live url list_type is `index`, then you should not use `index-timeshift_abs`, just `timeshift_abs`. But when in live channel URL is something else (like video/mono) then you should use that list_type, example `video-timeshift_abs` This is specific to `timeshift_abs` and `timeshift_rel` API. For `timeshift_rel`, when list_type is index then `rel-timeshift_rel` should be used. When something else (like video/mono) then as usual

phunkyfish:
Maybe simpler to just use timehift_abs.

Or is there any reason to use timeshift_rel?

Is timeshift_rel just using an offset?

Mr-Groch:
It should work both the same way (the same continuous stream should be sended), but we need to test this. If timeshift_abs will work OK, then we can use just that. There is some notice in flussnic docs about timeshift_abs:

> When you request HLS URL on a specific channel, Flussonic starts a new session.
> If you use a timeshift URL, any additional requests use the same existing session.
> All video requests runs relative to this existing session.
> So if you use the same time in timeshift_abs for multiple requests, really it's not pure "absolute" time, it's still related to the current session.
> Therefore every time you request the same time, you get a different video chunk.
> It's normal behavior and it's the only good way to implement HLS timeshift.

But I think we should still need to use old way (list_type-{utc}-{duration}) for playing from EPG. New way just for timeshifting from live. Another quote from doc, why:

> Be careful, if you request an URL for 24 hours, Flussonic will return a huge playlist. We saw a case where a client blocked a 100 megabit channel by a request to the same event playlist repeating several times per second. In this case, Flussonic returned a huge playlist of a few hundred kilobytes without much load due to well-tuned implementation of the archive.

phunkyfish:
That’s tricky as the inputstream addon only supports A single catchup URL format currently.

Plus a user will only select playback as live and just assume everything will just work.

But we should be able to test this with the TS stream for flussonic as it was always using a timeshift abs call. Just play something from 24 hours ago and see does it saturate your connection.

EDIT: in latest update m3u8 should work now the same as TS.

phunkyfish:
If we had to use two different api calls one for live (or close to live) and the other for archive where would we draw the line? I.e. what would we consider the correct time period to use the live api call?

The inputstream addon won't know if it's the currently live programme but will know how close to `now` it is. So we could pass an optional catchup `Live` URL to use for this purpose.

EDIT: The assumption here is that only m3u8 requires this feature. TS streams should not require this.

Mr-Groch:
Just my quick thought - if amount of seconds to rewind is smaller than current show total duration then "live" timeshift should be used. 

EDIT: Also I suggest to not use `now` as end time, for catchup modes with end time or duration placeholders (like xc or shift). Flussonic will cut duration exceeding `now `, but other backends may not. We should use current show end time, even if this time is in future. IMHO this could fix stream stopping problem when rewinding close to live  for other backends.

I'm currently trying to make contact with XC team member to discuss how to deal with this problem in XC

phunkyfish:
We won’t always know the programme duration. A user can use previous and next from kodi and the inputstream addon just gets a seek call. In fact now the addon just assumes a programme duration of 4 hours ignoring the actual EPG value. Good idea with limiting it by `now`.

So let’s use 4 hours as the cutoff between live and archive for the moment.

Mr-Groch:
I have found an issue with settings priority. It should be from least important to most important:
1. `timeshift` tag from m3u
2. Addon setting **All channels support catchup using mode** 
3. `catchup` tag from m3u

I have now list with timeshift="5" and no other tags, but this is flussonic backend. As I have said few days earlier - there are providers that uses tags incorrectly (this provider has 2 list types - pure flussonic and for SIPTV, but he is putting timeshift tag to both)...

phunkyfish:
> I have found an issue with settings priority. It should be from least important to most important:
> 
> 1. `timeshift` tag from m3u
> 2. Addon setting **All channels support catchup using mode**
> 3. `catchup` tag from m3u
> 
> I have now list with timeshift="5" and no other tags, but this is flussonic backend. As I have said few days earlier - there are providers that uses tags incorrectly (this provider has 2 list types - pure flussonic and for SIPTV, but he is putting timeshift tag to both)...

So you want the all channels support using mode to override the timeshift tag but never override a catchup tag?

There will be a point where the order will conflict no matter what we do. What about a provider that uses timeshift correctly and uses the all channels settings correctly to support xc? What’s required here may be another setting that ignore timeshift tag when using the all channels setting.

Mr-Groch:
This situation is not possible - if provider use timeshift tag correctly then it must be siptv catchup standard - it could not be proper xc list at the same time :) The order I gave IMO will cover most cases. `Timeshift` tag is obsolete and as I have seen - sometimes incorectly used, for this reason I think that should have minimal priority if other settings are set. If there will be `catchup` tag in m3u, then we can assume that provider know what he is doing and this tag should have most priority.

Back to the flussonic timeshift. I am testing new commit with flussonic HLS. Unfortunately it does not work very well. When I timeshift backward from live short period of time, often flussonic wil serve stream from the same point of time. Moreover when I stop and go back to the same channel - it will not play live, but still from the same point when I have stopped. I think this is because of mentioned i in earlier post flussonic sessions behaviour.

It looks like that:
1. I'm watching live and then try to seek 1 minute backward
2. Timeshift was succesfull. Now I want to seek more 30 second back.
3. This time seek is unsuccesfull - it will go back to the same point of time as in point 2.
4. I can try many times to seek backwards and even forwards 30 second but it will be unsuccesfull - still will be getting to the same point of time as in ppoint 2.
5. Now I'm pressing stop and play again the same channel -still instead of live stream, I will be watching from the same point of time as in point 2 :)

Maybe lets try with `timeshift_rel` this time?

for `index` list_type its:

`timeshift_rel-{offset}`

for other list_types:

`list_type-timeshift_rel-{offset}`

full example:

`http://list.tv:8888/325/mono-timeshift_rel-380.m3u8?token=secret`

phunkyfish:
> This situation is not possible - if provider use timeshift tag correctly then it must be siptv catchup standard - it could not be proper xc list at the same time :) The order I gave IMO will cover most cases. `Timeshift` tag is obsolete and as I have seen - sometimes incorectly used, for this reason I think that should have minimal priority if other settings are set. If there will be `catchup` tag in m3u, then we can assume that provider know what he is doing and this tag should have most priority.

Currently the timeshift tag will only be used if there is no catchup tag present. Are you saying that the all channels override is more important than the timeshift tag?

Mr-Groch:
In my opinion yes. But most imortant should be catchup tag if present. 

EDIT: I have other idea - `timeshift` tag if present, and there's no `catchup` tag and **All channels support catchup using mode** is disabled, then we should assume that this channel is SIPTV (shift). But when there is `timeshift` tag with no `catchup` tag and **All channels support catchup using mode** is set - then catchup should be enabled on channels with `timeshift` tag with type set in **All channels support catchup using mode**. If there is `timeshift` tag and `catchup` tag - then catchup on that channel should be enabled with type from `catchup` tag (**All channels support catchup using mode** in this situation is ignored).

phunkyfish:
> In my opinion yes. But most imortant should be catchup tag if present

Ok, the catchup tag is the highest priority currently. In the current implementation the timeshift tag is converted to a shift tag. So it would be a higher priority than the all channel using mode setting. It’s a bit of work to change that but if that works best I guess it can be done like that.

If the timeshift tag is the lowest priority should the catchup days it represents also be ignored?

Mr-Groch:
> If the timeshift tag is the lowest priority should the catchup days it represents also be ignored?

If there is `catchup-days` tag - then yes. But I think that `x` from `timeshift="x"` should have more priority than **Catchup window** from settings. Why? Becouse provider inserted that number, and we can assume that number is correct. So - days from `timeshift="x"` will have high priority (but lower than `catchup-days`), but catchup type is lowest prioryty - we can assume it is **SIPTV shift** only when no other option override that

Yes, I know it is getting complicated, but I see big potential in this addon. We can make something very universal here. Until now I have used many IPTV players with many providers - always there was need to use special list for each player, override some setings or even there was need to manually edit m3u lists.

Mr-Groch:
> Back to the flussonic timeshift. I am testing new commit with flussonic HLS. Unfortunately it does not work very well. When I timeshift backward from live short period of time, often flussonic wil serve stream from the same point of time. Moreover when I stop and go back to the same channel - it will not play live, but still from the same point when I have stopped. I think this is because of mentioned i in earlier post flussonic sessions behaviour.
> 
> It looks like that:
> 
> 1. I'm watching live and then try to seek 1 minute backward
> 2. Timeshift was succesfull. Now I want to seek more 30 second back.
> 3. This time seek is unsuccesfull - it will go back to the same point of time as in point 2.
> 4. I can try many times to seek backwards and even forwards 30 second but it will be unsuccesfull - still will be getting to the same point of time as in ppoint 2.
> 5. Now I'm pressing stop and play again the same channel -still instead of live stream, I will be watching from the same point of time as in point 2 :)
> 
> Maybe lets try with `timeshift_rel` this time?
> 
> for `index` list_type its:
> 
> `timeshift_rel-{offset}`
> 
> for other list_types:
> 
> `list_type-timeshift_rel-{offset}`
> 
> full example:
> 
> `http://list.tv:8888/325/mono-timeshift_rel-380.m3u8?token=secret`

I have tested also TS list. In this case all is OK - timeshift_abs for **Flussonic** TS works great. All - playing from EPG is OK, seeking in catchup is OK, timeshift from live is OK, stream will not stop when rewind small amount of time from live. It is my new favourite list now for using with catchup :)

phunkyfish:
Ok, I will try the timeshift_rel tomorrow. Will post back when it’s updated

phunkyfish:
Ok, latest version uses `timeshift_rel-{offset:1}` for m3u8. Let me know if it works. From the API docs it sounds like it should.

Mr-Groch:
After quick tests, it looks ways batter than before. There is no "bound to session" effect when trying to seek back/forwards a little amount of time

phunkyfish:
Ok, let me know if you are happy with this and I’ll clean up the code.

So that means outstanding would be:
 - The same for xc 
 - Capping duration to now 
 - Handling redirects, although that may get fixed by capping to now. I’m really not sure this has anything to do with redirects.

phunkyfish:
Do you also want to try this style of URL: `http://flussonic:8080/channel/index-1350274200-now.m3u8`

I think it's purpose is similar to duration but unbounded. However `timeshift_rel` is far simpler.

Mr-Groch:
I have tried that catchup URL in VLC - it will stop when play time reaches `now` - time of making request. So it's the same as `http://flussonic:8080/channel/index-{utc}-{duration}.m3u8` where `{duration}` is time from `{utc}` to `now` from addon.

To the TODO list I wish also add:
1. new `timeshift` tag behaviour

> But I think that `x` from `timeshift="x"` should have more priority than **Catchup window** from settings. Why? Because provider inserted that number, and we can assume that number is correct. So - days from `timeshift="x"` will have high priority (but lower than `catchup-days`), but catchup type from `timeshift` tag is lowest prioryty - we can assume it is **SIPTV shift** only when no other option override that (`catchup` tag or **All channels support catchup using mode** setting).
2. Restarting stream from catchup URL after resuming from pause when was watching live. Now when resuming from pause few seconds will play OK from buffer, but after that stream will jump to current live

phunkyfish:
Ok, will do those but might move pause issue and redirects to next release. There is a lot in this one already.

phunkyfish:
Latest ffmpegdirect should now have duration capped. Do you want to check if that has any effect on the redirects issue?

Mr-Groch:
Have tried latest ffmpegdirect with shift catchup type list that redirects to other stream type when trying to timeshift. It didn't fix the problem... Still when trying to timeshift from live, video and audio freezes and KODI can't play but it keeps trying...

Maybe this will help what I mean by redirects - this provider serves this video.m3u8 when playing live:
```
#EXTM3U
#EXT-X-TARGETDURATION:6
#EXT-X-VERSION:3
#EXT-X-MEDIA-SEQUENCE:26007
#EXT-X-PROGRAM-DATE-TIME:2020-02-22T13:24:41Z
#EXTINF:5.120,
2020/02/22/13/24/41-05120.ts?token=secret
#EXTINF:5.600,
2020/02/22/13/24/46-05600.ts?token=secret
#EXTINF:5.920,
2020/02/22/13/24/51-05920.ts?token=secret
#EXTINF:5.120,
2020/02/22/13/24/57-05120.ts?token=secret

```

But when accesing catchup URL - it will serve HLS index:
```
#EXTM3U
#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="aac",NAME="pol a1",DEFAULT=YES,AUTOSELECT=YES,LANGUAGE="pol"
#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="aac",NAME="pol a2",DEFAULT=NO,AUTOSELECT=YES,LANGUAGE="pol",URI="tracks-a2/index-1582377046-3600.m3u8?token=secret"
#EXT-X-STREAM-INF:AUDIO="aac",CLOSED-CAPTIONS=NONE,RESOLUTION=1920x1080,FRAME-RATE=25.000,AVERAGE-BANDWIDTH=5170000,BANDWIDTH=6470000
tracks-v1a1/index-1582377046-3600.m3u8?token=secret
```
When I play catchup from EPG - this is working, seeking backward/forward also. But when playing live and try ti timeshift - then it freezes... As I had said in mail - timeshifting is working in Archive Client addon.

phunkyfish:
That looks correct for live. It should keep serving new TS files continuously.

Did you expect something different for live?

Mr-Groch:
Both are correct. When watching live it will serve new TS files continuously. But when I try to timeshift backward from live then backend will serve that second index.m3u8 and player can't handle with this. I think that player expects same stream as live, but he get HLS index with link to other index where ts segments are...

phunkyfish:
Ok, `timeshift` tag priorities added. Let's move the remaining items to new github issues and move to getting this PR merged.

Mr-Groch:
OK, I will create new issues for remaining TODO's. But it looks that `timeshift` tag priority doesn't work as it should. I  have tested list with `timeshift="5"` tag, no other catchup tags, and in addon settings I have **All channels support catchup using mode** = **flussonic**. When I try to play from catchup - shift cacthup type is used, not flussonic.

phunkyfish:
> OK, I will create new issues for remaining TODO's. But it looks that `timeshift` tag priority doesn't work as it should. I have tested list with `timeshift="5"` tag, no other catchup tags, and in addon settings I have **All channels support catchup using mode** = **flussonic**. When I try to play from catchup - shift cacthup type is used, not flussonic.

Ok, that’s a bug.

phunkyfish:
Ok, fixed now @Mr-Groch

Mr-Groch:
Fixed, confirmed, thanks :)

AlwinEsch:
Related to https://github.com/xbmc/xbmc/pull/17307, not forget a rebase of request to allow further Jenkins use!

```sh
git remote add upstream https://github.com/kodi-pvr/pvr.iptvsimple
git fetch upstream
git rebase upstream/Matrix
```

PR #346, opened 2020-03-30
Fix incorrect live URL selection for standard (no timeshift) catchup streams

phunkyfish:
v4.11.6
- Fixed: Fix incorrect live URL selection and inputstream for standard (no timeshift) catchup streams
- Fixed: Correctly cache mime_type for streams

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ Settings related to Channel Logos.
 ### Catchup
 Catchup settings for viewing archived live programmes. Allows the option of 'Play programme' when viewing EPG entry info. Only works if your provider supports catchup.
 
-Note that catchup functionality can work one of two ways. The first is where a standard inputstream is used, allowing the use of the EPG to view a programme from within the catchup window like a video. The second is where the `inputstream.ffmpegdirect` addon is installed which allows timeshift, either as a video or with playback as live. When playback as live is enabled it's possible to skip back/forward programme by programme and jump to other channel and programmes via the OSD menus.
+Note that catchup functionality can work one of two ways. The first is where a standard inputstream is used, allowing the use of the EPG to view a programme from within the catchup window like a video. The second is where the `inputstream.ffmpegdirect` addon is installed which allows timeshift, either as a video or with playback as live. When playback as live is enabled it's possible to skip back/forward programme by programme and jump to other channel and programmes via the OSD menus. Also note that using `inputstream.ffmpegdirect` is automatic if using TS/M3U8 streams that support catchup. If this inputstream is incorrectly chosen it's possible to override this by adding a `#KODIPROP:inputstreamclass=inputstream.ffmpeg` line before the channel's M3U entry (or any alternative inputstream instead).
 
 Catchup tags can be specified in the M3U entries and these tags override any values configured in the Catchup settings. The supported M3U catchup tags are `catchup`, `catchup-source` and `catchup-days`. See the [Supported M3U elements section](#m3u-format-elements) below for further details. The full catchup URL or a query to be appended to the stream URL is either provided in the `catchup-source` tag (i.e. complete with format specifiers) or in the case this is not provided the `Query format string` setting is appended to stream URL. See the [Catchup format specifiers section](#catchup-format-specifiers) below for a more detailed explanation and further examples.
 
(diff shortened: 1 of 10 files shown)


phunkyfish:
@Sorien

Could you test this? It will only effect standard catchup streams that don't use `inputstream.ffmpegdirect`. But no harm in double checking.

Sorien:
@phunkyfish can i find somewhere artifacts from win64 build?

edit:
not sure if i'll be able to test anything, I just downloaded lastest kodi from http://mirrors.kodi.tv/nightlies/windows/win64/master/ 
KodiSetup-20200329-e7878130-master-x64.exe 

and i can't play any stream from channel list or epg kodi crashes without any logged error it's same as last time and pretty i'm sure that im not alone, i was talking to one guy last weekend and he had same problem

edit2:
~if you could find some version with pdb file i'll attach it to debuger~
i have found version with pdb (http://mirrors.kodi.tv/test-builds/windows/win64/) KodiSetup-20200329-cd865190-PR17576-merge-x64.exe

Unhandled exception at 0x00007FFCC5FA2590 (ucrtbase.dll) in kodi.exe: 0xC00000FD: Stack overflow (parameters: 0x0000000000000001, 0x000000F07CE33FE8). occurred

https://gist.github.com/phunkyfish/cec957e24e1eccdf1595b8931df36531

phunkyfish:
I updated your crash log with a gist ;)

You would also need to use the latest addon version with that kodi as there would have been PVR API changes.

Latest addon build is here: https://jenkins.kodi.tv/blue/organizations/jenkins/kodi-pvr%2Fpvr.iptvsimple/detail/PR-346/2/artifacts

phunkyfish:
@Mr-Groch @SerhiyRomanov

Could you both check if timeshifting is still ok?

Mr-Groch:
I have checked `timeshift="x"` (SIPTV) catchup with HLS and MPEGTS - for both timeshift works with this PR

SerhiyRomanov:
@phunkyfish Are anywhere in Jenkins the build for Ubuntu? Or should I build it every time by myself?

phunkyfish:
I think you would need to build yourself. I think with ubuntu there are just too many options.

SerhiyRomanov:
Sorry for the late response,  I've just checked it for Flussonic - everything works as before.

puenktchen:
@phunkyfish the problem is somehow back in the nightlies.
After playing a channel with "Play Program" (catchup-url), every following "Switch" on the same or other channel also plays the last catchup-url instead of the actual channel-url.
In the log: [https://paste.kodi.tv/olutosedic.kodi](url) i've started with "Play Program", stopped playback and then "Switched" to an other channel.

phunkyfish:
Your log link does not work.

Also, what platform are you on?

puenktchen:
I'm testing this on Windows 10. Here's the corrected link:
https://paste.kodi.tv/olutosedic.kodi

To precise my last post. If i start live tv by clicking on a channel in the channel list, all is fine. I can switch from one channel to an other and the pvr addon always uses the live channel url. But after calling a catchup stream with the "Play Programm" action the problem occurs. It's like the catchup url is cached somehow and doesn't get cleared for channel "Switch" actions.

phunkyfish:
And are you using timeshifted catchup or not?

Maybe create an issue for this so we are not spamming the others on this PR thread.

puenktchen:
No timeshifted catchup! Just a simple catchup-url with catchup-id as the only parameter.
I'm the one who initiated all of this, the one who gave you access to my server in a private conversation
of the Kodi forum. You remember?

phunkyfish:
Ok, I remember now :)

Can you give me access again and I will try and fix tomorrow? Just PM me details on the kodi forum.

puenktchen:
Will do in a few hours.

phunkyfish:
@puenktchen 

Do you have "Play from EPG in live TV mode (using timeshift)" enabled or disabled? If not using timeshift this should be disabled. Although if having this enabled causes this issue that is a bug anyway.

puenktchen:
Will do in a few hours.

> 
> 
> @puenktchen
> 
> Do you have "Play from EPG in live TV mode (using timeshift)" enabled or disabled? If not using timeshift this should be disabled. Although if having this enabled causes this issue that is a bug anyway.

Yes, the mentioned setting was enabled on my side. Don't know why i did this.

PR #352, opened 2020-04-14
Provide media headers to inputstream.adaptive if channel does not explicitly set them.

basilgello:
Fixes the HTTP 403 error on segment download with several
IPTV providers, as suggested by Markus Pfau (@peak3d)
here: https://github.com/peak3d/inputstream.adaptive/issues/411

### DIFF
--- a/src/iptvsimple/utilities/StreamUtils.cpp
+++ b/src/iptvsimple/utilities/StreamUtils.cpp
@@ -61,9 +61,35 @@ void StreamUtils::SetAllStreamProperties(PVR_NAMED_VALUE* properties, unsigned i
           StreamUtils::SetStreamProperty(properties, propertiesCount, propertiesMax, PVR_STREAM_PROPERTY_INPUTSTREAMCLASS, PVR_STREAM_PROPERTY_VALUE_INPUTSTREAMFFMPEG);
       }
     }
-    else // inputstream.adpative
+    else // inputstream.adaptive
     {
-      StreamUtils::SetStreamProperty(properties, propertiesCount, propertiesMax, PVR_STREAM_PROPERTY_STREAMURL, streamURL);
+      bool streamUrlSet = false;
+
+      // If no media headers are explicitly set for inputstream.adaptive,
+      // strip the headers from streamURL and put it to media headers property
+
+      if (channel.GetProperty("inputstream.adaptive.stream_headers").empty())
+      {
+        // No stream headers declared by property, check if stream URL has any
+        size_t found = streamURL.find_first_of('|');
+        if (found != std::string::npos)
+        {
+          // Headers found, split and url-encode them
+          const std::string& url = streamURL.substr(0, found);
+          const std::string& protocolOptions = streamURL.substr(found + 1, streamURL.length());
+          const std::string& encodedProtocolOptions = StreamUtils::GetUrlEncodedProtocolOptions(protocolOptions);
+
+          // Set stream URL without headers and encoded headers as property
+          StreamUtils::SetStreamProperty(properties, propertiesCount, propertiesMax, PVR_STREAM_PROPERTY_STREAMURL, url);
+          StreamUtils::SetStreamProperty(properties, propertiesCount, propertiesMax, "inputstream.adaptive.stream_headers", encodedProtocolOptions);
+          streamUrlSet = true;
+        }
+      }
+
+      // Set intact stream URL if not previously set
+      if (!streamUrlSet)
+        StreamUtils::SetStreamProperty(properties, propertiesCount, propertiesMax, PVR_STREAM_PROPERTY_STREAMURL, streamURL);
+
       StreamUtils::SetStreamProperty(properties, propertiesCount, propertiesMax, PVR_STREAM_PROPERTY_INPUTSTREAMCLASS, "inputstream.adaptive");
       StreamUtils::SetStreamProperty(properties, propertiesCount, propertiesMax, "inputstream.adaptive.manifest_type", StreamUtils::GetManifestType(streamType));
       if (streamType == StreamType::HLS || streamType == StreamType::DASH)
@@ -214,29 +240,34 @@ std::string StreamUtils::GetURLWithFFmpegReconnectOptions(const std::string& str
 
 std::string StreamUtils::AddHeaderToStreamUrl(const std::string& streamUrl, const std::string& headerName, const std::string& headerValue)
 {
-  std::string newStreamUrl = streamUrl;
+  return StreamUtils::AddHeader(streamUrl, headerName, headerValue, false);
+}
+
+std::string StreamUtils::AddHeader(const std::string& headerTarget, const std::string& headerName, const std::string& headerValue, bool encodeHeaderValue)
+{
+  std::string newHeaderTarget = headerTarget;
 
   bool hasProtocolOptions = false;
   bool addHeader = true;
-  size_t found = newStreamUrl.find("|");
+  size_t found = newHeaderTarget.find("|");
 
   if (found != std::string::npos)
   {
     hasProtocolOptions = true;
-    addHeader = newStreamUrl.find(headerName + "=", found + 1) == std::string::npos;
+    addHeader = newHeaderTarget.find(headerName + "=", found + 1) == std::string::npos;
   }
 
   if (addHeader)
   {
     if (!hasProtocolOptions)
-      newStreamUrl += "|";
+      newHeaderTarget += "|";
     else
-      newStreamUrl += "&";
+      newHeaderTarget += "&";
 
-    newStreamUrl += headerName + "=" + headerValue;
+    newHeaderTarget += headerName + "=" + (encodeHeaderValue ? WebUtils::UrlEncode(headerValue) : headerValue);
   }
 
-  return newStreamUrl;
+  return newHeaderTarget;
 }
 
 bool StreamUtils::UseKodiInputstreams(const StreamType& streamType)
@@ -255,4 +286,25 @@ bool StreamUtils::SupportsFFmpegReconnect(const StreamType& streamType, const ip
   return streamType == StreamType::HLS ||
          channel.GetProperty(PVR_STREAM_PROPERTY_INPUTSTREAMCLASS) == PVR_STREAM_PROPERTY_VALUE_INPUTSTREAMFFMPEG ||
          channel.GetProperty(PVR_STREAM_PROPERTY_INPUTSTREAMADDON) == PVR_STREAM_PROPERTY_VALUE_INPUTSTREAMFFMPEG;
+}
+
+std::string StreamUtils::GetUrlEncodedProtocolOptions(const std::string& protocolOptions)
+{
+  std::string encodedProtocolOptions = "";
+
+  std::vector<std::string> headers = StringUtils::Split(protocolOptions, "&");
+  for (std::string header : headers)
+  {
+    std::string::size_type pos(header.find('='));
+    if(pos == std::string::npos)
+      continue;
+
+    encodedProtocolOptions = StreamUtils::AddHeader(encodedProtocolOptions, header.substr(0, pos), header.substr(pos + 1), true);
+  }
+
+  // We'll return the protocol options without the leading '|'
+  if (!encodedProtocolOptions.empty() && encodedProtocolOptions[0] == '|')
+    encodedProtocolOptions.erase(0, 1);
+
+  return encodedProtocolOptions;
 }
--- a/src/iptvsimple/utilities/StreamUtils.h
+++ b/src/iptvsimple/utilities/StreamUtils.h
@@ -41,10 +41,11 @@ namespace iptvsimple
       static const std::string GetManifestType(const StreamType& streamType);
       static const std::string GetMimeType(const StreamType& streamType);
       static std::string GetURLWithFFmpegReconnectOptions(const std::string& streamUrl, const StreamType& streamType, const iptvsimple::data::Channel& channel);
+      static std::string AddHeader(const std::string& headerTarget, const std::string& headerName, const std::string& headerValue, bool encodeHeaderValue);
       static std::string AddHeaderToStreamUrl(const std::string& streamUrl, const std::string& headerName, const std::string& headerValue);
       static bool UseKodiInputstreams(const StreamType& streamType);
       static bool ChannelSpecifiesInputstream(const iptvsimple::data::Channel& channe);
-
+      static std::string GetUrlEncodedProtocolOptions(const std::string& protocolOptions);
       static std::string GetEffectiveInputStreamClass(const StreamType& streamType, const iptvsimple::data::Channel& channel);
 
     private:
--- a/pvr.iptvsimple/addon.xml.in
+++ b/pvr.iptvsimple/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.iptvsimple"
-  version="4.11.6"
+  version="4.11.7"
   name="PVR IPTV Simple Client"
   provider-name="nightik">
   <requires>@ADDON_DEPENDS@</requires>
@@ -165,6 +165,9 @@
       <icon>icon.png</icon>
     </assets>
     <news>
+v4.11.7
+- Fixed: stream media headers not set when inputstream.adaptive is used for HLS streams
+
 v4.11.6
 - Fixed: Fix incorrect live URL selection and inputstream for standard (no timeshift) catchup streams
 - Fixed: Correctly cache mime_type for streams
--- a/pvr.iptvsimple/changelog.txt
+++ b/pvr.iptvsimple/changelog.txt
@@ -1,3 +1,6 @@
+v4.11.7
+- Fixed: stream media headers not set when inputstream.adaptive is used for HLS streams
+
 v4.11.6
 - Fixed: Fix incorrect live URL selection and inputstream for standard (no timeshift) catchup streams
 - Fixed: Correctly cache mime_type for streams


src/iptvsimple/utilities/StreamUtils.cpp
```diff
@@ -63,7 +63,41 @@ void StreamUtils::SetAllStreamProperties(PVR_NAMED_VALUE* properties, unsigned i
     }
     else // inputstream.adpative
     {
-      StreamUtils::SetStreamProperty(properties, propertiesCount, propertiesMax, PVR_STREAM_PROPERTY_STREAMURL, streamURL);
+      // If no media headers are explicitly set for inputstream.adaptive,
+      // strip the headers from streamURL and put it to media headers property
+
+      if (channel.GetProperty("inputstream.adaptive.stream_headers").empty()) {
```
phunkyfish: Brackets on the following line please. The same as the rest of the code.

src/iptvsimple/utilities/StreamUtils.cpp
```diff
@@ -63,7 +63,41 @@ void StreamUtils::SetAllStreamProperties(PVR_NAMED_VALUE* properties, unsigned i
...
+        std::string url = streamURL;
+        std::string protocolOptions;
+        std::string untrimmed_header = "";
+
+        size_t found = streamURL.find_first_of('|');
+        if (found != std::string::npos) {
+          url = streamURL.substr(0, found);
+          protocolOptions = streamURL.substr(found + 1, streamURL.length());
+
+          std::vector<std::string> headers = StringUtils::Split(protocolOptions, "&");
```
phunkyfish: This is very confusing. Are you stripping off and splitting each header only to add them all back the same way? If so why?

If you did need to add headers again there is a function for that already but I don't think it's needed at all. 
https://github.com/kodi-pvr/pvr.iptvsimple/blob/48ace1a75a72ff83e7baea7d23ea90bfc2334df6/src/iptvsimple/utilities/StreamUtils.cpp#L215-L240

src/iptvsimple/utilities/StreamUtils.cpp
```diff
@@ -63,7 +63,41 @@ void StreamUtils::SetAllStreamProperties(PVR_NAMED_VALUE* properties, unsigned i
...
+          }
+
+          protocolOptions.pop_back();
+
+          StreamUtils::SetStreamProperty(properties, propertiesCount, propertiesMax, PVR_STREAM_PROPERTY_STREAMURL, url);
+          StreamUtils::SetStreamProperty(properties, propertiesCount, propertiesMax, "inputstream.adaptive.stream_headers", protocolOptions);
+        } else {
+          StreamUtils::SetStreamProperty(properties, propertiesCount, propertiesMax, PVR_STREAM_PROPERTY_STREAMURL, streamURL);
+        }
+      } else
```
phunkyfish: If the rest of the if/else has brackets don't leave one hanging like this ;)

Also, I don't think you need this line repeated. Just use a bool to see if the stream_headers were set and then only do this one if they were not.

phunkyfish (changes requested):
Ok, I think you are doing more than you need to with this. It looks like you strip the headers to just add them in the same way again. Can you give some examples of URLs and what the value of `stream_headers` ends up being?

src/iptvsimple/utilities/StreamUtils.cpp
basilgello: I am splitting the header and url-encoding the value to match https://github.com/peak3d/inputstream.adaptive/blob/Matrix/src/helpers.cpp#L459

Makes sense to upgrade the function you mentioned with "urlencode" boolean parameter, though.

src/iptvsimple/utilities/StreamUtils.cpp
basilgello: > Can you give some examples of URLs and what the value of `stream_headers` ends up being?

Let me start the test VM up and send you the debug log by email.

src/iptvsimple/utilities/StreamUtils.cpp
basilgello: done

src/iptvsimple/utilities/StreamUtils.cpp
basilgello: done

src/iptvsimple/utilities/StreamUtils.cpp
phunkyfish: Ya, you should upgrade it 😉

You could also add another function to StreamUtils to encode the protocol options and that would keep this function clean and allow the url encoring be used by other inputstreams in the future.

src/iptvsimple/utilities/StreamUtils.cpp
```diff
@@ -63,7 +63,56 @@ void StreamUtils::SetAllStreamProperties(PVR_NAMED_VALUE* properties, unsigned i
...
+
+        size_t found = streamURL.find_first_of('|');
+        if (found != std::string::npos)
+        {
+          url = streamURL.substr(0, found);
+          protocolOptions = streamURL.substr(found + 1, streamURL.length());
+
+          std::vector<std::string> headers = StringUtils::Split(protocolOptions, "&");
+          protocolOptions.clear();
+          for (std::vector<std::string>::iterator b(headers.begin()), e(headers.end()); b != e; ++b)
```
phunkyfish: why not just `for (std::string header : headers)` and then just find the `=` in each header?

Much easier to read and understand.

src/iptvsimple/utilities/StreamUtils.cpp
```diff
@@ -63,7 +63,56 @@ void StreamUtils::SetAllStreamProperties(PVR_NAMED_VALUE* properties, unsigned i
...
+          {
+            std::string::size_type pos(b->find('='));
+            if(pos == std::string::npos)
+            {
+              continue;
+            }
+
+            untrimmed_header = b->substr(0, pos);
+            protocolOptions += StringUtils::Trim(untrimmed_header);
+            untrimmed_header = b->substr(pos+1);
```
phunkyfish: `pos + 1`

src/iptvsimple/utilities/StreamUtils.cpp
```diff
@@ -63,7 +63,56 @@ void StreamUtils::SetAllStreamProperties(PVR_NAMED_VALUE* properties, unsigned i
...
+            protocolOptions += "=" + WebUtils::UrlEncode(StringUtils::Trim(untrimmed_header)) + "&";
+          }
+
+          protocolOptions.pop_back();
```
phunkyfish: This is a bit hacky ;)

phunkyfish (changes requested):
Really you should update the AddHeader call to optionally support URL encoding. I think your code to encode the headers could still be simpler.

src/iptvsimple/utilities/StreamUtils.cpp
basilgello: I also thought of AddHeader modification in the meantime. Let me run it quickly and see if code gets simpler :)

src/iptvsimple/utilities/StreamUtils.cpp
basilgello: I added new function and tested the build.

src/iptvsimple/utilities/StreamUtils.cpp
```diff
@@ -63,7 +63,37 @@ void StreamUtils::SetAllStreamProperties(PVR_NAMED_VALUE* properties, unsigned i
...
     {
-      StreamUtils::SetStreamProperty(properties, propertiesCount, propertiesMax, PVR_STREAM_PROPERTY_STREAMURL, streamURL);
+      // If no media headers are explicitly set for inputstream.adaptive,
+      // strip the headers from streamURL and put it to media headers property
+
+      bool haveStreamHeaders = !channel.GetProperty("inputstream.adaptive.stream_headers").empty();
+      if (!haveStreamHeaders)
+      {
+        // preserve any kodi protocol options after "|"
+        std::string url = streamURL;
```
phunkyfish: Defining these here serves no purpose. They can be defined where are used. Also they should be const.

src/iptvsimple/utilities/StreamUtils.cpp
```diff
@@ -256,3 +286,24 @@ bool StreamUtils::SupportsFFmpegReconnect(const StreamType& streamType, const ip
          channel.GetProperty(PVR_STREAM_PROPERTY_INPUTSTREAMCLASS) == PVR_STREAM_PROPERTY_VALUE_INPUTSTREAMFFMPEG ||
          channel.GetProperty(PVR_STREAM_PROPERTY_INPUTSTREAMADDON) == PVR_STREAM_PROPERTY_VALUE_INPUTSTREAMFFMPEG;
 }
+
+std::string StreamUtils::UrlEncodeProtocolOptions(const std::string& protocolOptions)
```
phunkyfish: Maybe call the function `GetUrlEncodedProtocolOptions`?

Might be nice to all have a `HasProtocolOptions` function.

phunkyfish (changes requested):
I still think you should add a boolean to https://github.com/kodi-pvr/pvr.iptvsimple/blob/53c7c978aea9e512fe6b206c7a66daea575e637f/src/iptvsimple/utilities/StreamUtils.cpp#L245

And then do the URL encoding as part pr that function. You could change the name to something else if you wish. Like `AddHeader` and the argument just called a simple string name or maybe `headerTarget`.

src/iptvsimple/utilities/StreamUtils.cpp
```diff
@@ -63,7 +63,37 @@ void StreamUtils::SetAllStreamProperties(PVR_NAMED_VALUE* properties, unsigned i
...
+          StreamUtils::SetStreamProperty(properties, propertiesCount, propertiesMax, "inputstream.adaptive.stream_headers", encodedProtocolOptions);
+        }
+        else
+        {
+          // There are no headers passed - assign the url as is
+          haveStreamHeaders = true;
+        }
+      }
+
+      if (haveStreamHeaders)
```
phunkyfish: if (!haveStreamHeaders)

src/iptvsimple/utilities/StreamUtils.cpp
```diff
@@ -63,7 +63,37 @@ void StreamUtils::SetAllStreamProperties(PVR_NAMED_VALUE* properties, unsigned i
...
+      {
```
phunkyfish: No need for bracket

src/iptvsimple/utilities/StreamUtils.cpp
```diff
@@ -63,7 +63,37 @@ void StreamUtils::SetAllStreamProperties(PVR_NAMED_VALUE* properties, unsigned i
...
+        StreamUtils::SetStreamProperty(properties, propertiesCount, propertiesMax, PVR_STREAM_PROPERTY_STREAMURL, streamURL);
+      }
```
phunkyfish: No need for bracket

src/iptvsimple/utilities/StreamUtils.cpp
```diff
@@ -63,7 +63,37 @@ void StreamUtils::SetAllStreamProperties(PVR_NAMED_VALUE* properties, unsigned i
...
+          size_t found = streamURL.find_first_of('|');
+          const std::string& url = streamURL.substr(0, found);
+          const std::string& protocolOptions = streamURL.substr(found + 1, streamURL.length());
+          std::string encodedProtocolOptions = StreamUtils::GetUrlEncodedProtocolOptions(protocolOptions);
+
+          // inputstream.adaptive.stream_headers does not recognize first '|' so remove it
+          encodedProtocolOptions.erase(0, 1);
+
+          StreamUtils::SetStreamProperty(properties, propertiesCount, propertiesMax, PVR_STREAM_PROPERTY_STREAMURL, url);
+          StreamUtils::SetStreamProperty(properties, propertiesCount, propertiesMax, "inputstream.adaptive.stream_headers", encodedProtocolOptions);
```
phunkyfish, line 84: after this line `haveStreamHeaders = true;`

src/iptvsimple/utilities/StreamUtils.cpp
```diff
@@ -63,7 +63,37 @@ void StreamUtils::SetAllStreamProperties(PVR_NAMED_VALUE* properties, unsigned i
...
+        }
+        else
```
phunkyfish: Drop the whole else block ;)

src/iptvsimple/utilities/StreamUtils.cpp
```diff
@@ -63,7 +63,37 @@ void StreamUtils::SetAllStreamProperties(PVR_NAMED_VALUE* properties, unsigned i
...
     else // inputstream.adpative
     {
-      StreamUtils::SetStreamProperty(properties, propertiesCount, propertiesMax, PVR_STREAM_PROPERTY_STREAMURL, streamURL);
+      // If no media headers are explicitly set for inputstream.adaptive,
+      // strip the headers from streamURL and put it to media headers property
+
+      bool haveStreamHeaders = !channel.GetProperty("inputstream.adaptive.stream_headers").empty();
+      if (!haveStreamHeaders)
+      {
+        if (StreamUtils::HasProtocolOptions(streamURL))
```
phunkyfish: My bad. Providing this function doesn't add any efficiency. It just means you have to do the find twice. Sorry. So you should go back to the `size_t found =` above the if statement and found in the if.

Sorry again!

src/iptvsimple/utilities/StreamUtils.cpp
```diff
@@ -256,3 +291,28 @@ bool StreamUtils::SupportsFFmpegReconnect(const StreamType& streamType, const ip
...
+  {
+    std::string::size_type pos(header.find('='));
+    if(pos == std::string::npos)
+    {
+      continue;
+    }
+
+    encodedProtocolOptions = StreamUtils::AddHeader(encodedProtocolOptions, header.substr(0, pos), header.substr(pos + 1), true);
+  }
+
```
phunkyfish, line 304: Add this here instead.

```cpp
  // We'll return the protocol options without the leading '|'
  if (!encodedProtocolOptions.erase.empty())
    encodedProtocolOptions.erase(0, 1);

File: src/iptvsimple/utilities/StreamUtils.cpp [hunk 16]
```diff
@@ -63,7 +63,37 @@ void StreamUtils::SetAllStreamProperties(PVR_NAMED_VALUE* properties, unsigned i
...
+      if (!haveStreamHeaders)
+      {
+        if (StreamUtils::HasProtocolOptions(streamURL))
+        {
+          size_t found = streamURL.find_first_of('|');
+          const std::string& url = streamURL.substr(0, found);
+          const std::string& protocolOptions = streamURL.substr(found + 1, streamURL.length());
+          std::string encodedProtocolOptions = StreamUtils::GetUrlEncodedProtocolOptions(protocolOptions);
+
+          // inputstream.adaptive.stream_headers does not recognize first '|' so remove it
```
phunkyfish: This would be better in your new function I think.

src/iptvsimple/utilities/StreamUtils.cpp
```diff
@@ -256,3 +291,28 @@ bool StreamUtils::SupportsFFmpegReconnect(const StreamType& streamType, const ip
          channel.GetProperty(PVR_STREAM_PROPERTY_INPUTSTREAMCLASS) == PVR_STREAM_PROPERTY_VALUE_INPUTSTREAMFFMPEG ||
          channel.GetProperty(PVR_STREAM_PROPERTY_INPUTSTREAMADDON) == PVR_STREAM_PROPERTY_VALUE_INPUTSTREAMFFMPEG;
 }
+
+bool StreamUtils::HasProtocolOptions(const std::string& streamUrl)
+{
+  size_t found = streamUrl.find_first_of('|');
+  return (found != std::string::npos);
+}
```
phunkyfish: Just drop this guy, apologies again.

phunkyfish (changes requested):

src/iptvsimple/utilities/StreamUtils.cpp
```diff
@@ -214,29 +240,34 @@ std::string StreamUtils::GetURLWithFFmpegReconnectOptions(const std::string& str
 
 std::string StreamUtils::AddHeaderToStreamUrl(const std::string& streamUrl, const std::string& headerName, const std::string& headerValue)
 {
-  std::string newStreamUrl = streamUrl;
+  return StreamUtils::AddHeader(streamUrl, headerName, headerValue, false);
+}
+
+std::string StreamUtils::AddHeader(const std::string& headerTarget, const std::string& headerName, const std::string& headerValue, const bool encodeHeaderValue)
```
phunkyfish: You don't need the const for the bool as it's a basic type.

src/iptvsimple/utilities/StreamUtils.cpp
```diff
@@ -256,3 +287,26 @@ bool StreamUtils::SupportsFFmpegReconnect(const StreamType& streamType, const ip
...
+std::string StreamUtils::GetUrlEncodedProtocolOptions(const std::string& protocolOptions)
+{
+  std::string encodedProtocolOptions = "";
+
+  std::vector<std::string> headers = StringUtils::Split(protocolOptions, "&");
+  for (std::string header : headers)
+  {
+    std::string::size_type pos(header.find('='));
+    if(pos == std::string::npos)
+    {
```
phunkyfish: Don't need the bracket

src/iptvsimple/utilities/StreamUtils.cpp
```diff
@@ -256,3 +287,26 @@ bool StreamUtils::SupportsFFmpegReconnect(const StreamType& streamType, const ip
...
+      continue;
+    }
```
phunkyfish: Don't need the bracket

src/iptvsimple/utilities/StreamUtils.h
```diff
@@ -41,10 +41,11 @@ namespace iptvsimple
       static const std::string GetManifestType(const StreamType& streamType);
       static const std::string GetMimeType(const StreamType& streamType);
       static std::string GetURLWithFFmpegReconnectOptions(const std::string& streamUrl, const StreamType& streamType, const iptvsimple::data::Channel& channel);
+      static std::string AddHeader(const std::string& headerTarget, const std::string& headerName, const std::string& headerValue, const bool encodeHeaderValue);
```
phunkyfish: Same again, don't need the const for the bool

phunkyfish (changes requested):
Ok, this is looking really good. Just some minors.

You also need to update the addon.xml.in and changelog.txt. There are chagnelog entries in both files.

For the version number you need to update minor number. I.e. in major.minor.patch the middle one as you've added a new feature.

Well done ;)

basilgello:
Technically it is a bugfix so I changed the patch instead of minor :) With timeshift occupying 4.12.0 in the PRs list, would be good to make it live ;)

Thanks for review! :)

phunkyfish (approved):
LGTM

phunkyfish:
And thanks for contributing!

PR #365, opened 2020-04-25
[Matrix] change to new C++ PVR interface way

AlwinEsch:
Changes related to https://github.com/xbmc/xbmc/pull/16485.

This make the addon to the new C++ addon interface.


src/iptvsimple/Settings.cpp
```diff
@@ -23,192 +22,153 @@ void Settings::ReadFromAddon(const std::string& userPath, const std::string clie
...
-    return SetSetting<int, ADDON_STATUS>(settingName, settingValue, m_startChannelNumber, ADDON_STATUS_OK, ADDON_STATUS_OK);
-  if (settingName == "numberByOrder")
-    return SetSetting<bool, ADDON_STATUS>(settingName, settingValue, m_numberChannelsByM3uOrderOnly, ADDON_STATUS_OK, ADDON_STATUS_OK);
-  if (settingName == "m3uRefreshMode")
-    return SetSetting<RefreshMode, ADDON_STATUS>(settingName, settingValue, m_m3uRefreshMode, ADDON_STATUS_OK, ADDON_STATUS_OK);
-  if (settingName == "m3uRefreshIntervalMins")
-    return SetSetting<int, ADDON_STATUS>(settingName, settingValue, m_m3uRefreshIntervalMins, ADDON_STATUS_OK, ADDON_STATUS_OK);
-  if (settingName == "m3uRefreshHour")
-    return SetSetting<int, ADDON_STATUS>(settingName, settingValue, m_m3uRefreshHour, ADDON_STATUS_OK, ADDON_STATUS_OK);
+    m_m3uPathType = static_cast<PathType>(settingValue.GetInt());
```
phunkyfish: This is reading the value and not setting it.

src/iptvsimple/utilities/FileUtils.cpp
```diff
@@ -142,15 +140,19 @@ int FileUtils::GetCachedFileContents(const std::string& cachedName, const std::s
...
+    STAT_STRUCTURE statCached;
+    STAT_STRUCTURE statOrig;
 
-    XBMC->StatFile(cachedPath.c_str(), &statCached);
-    XBMC->StatFile(filePath.c_str(), &statOrig);
+    kodi::vfs::StatFile(cachedPath, statCached);
+    kodi::vfs::StatFile(filePath, statOrig);
 
-    needReload = statCached.st_mtime < statOrig.st_mtime || statOrig.st_mtime == 0;
+#ifdef TARGET_WINDOWS
```
phunkyfish: Do we really need an ifdef here? Can kodi not do the OS specific stuff internally?

src/iptvsimple/Settings.h
```diff
@@ -129,38 +129,6 @@ namespace iptvsimple
     Settings(Settings const&) = delete;
     void operator=(Settings const&) = delete;
```
phunkyfish, line 136: After the set setting is fixed lets see what it looks like. The purpose this template was to tidy of the setting code and to always log setting changes.

AlwinEsch: Can bring back a template to do log them, NP.

src/iptvsimple/utilities/FileUtils.cpp
AlwinEsch: Really not nice, thought also about.
I look to make it better, currently is the `stat` on interface to much OS dependent.

src/iptvsimple/Settings.cpp
AlwinEsch: Not understand what you mean?

AlwinEsch:
Is updated and test was OK, as part of filesystem it need to have the other request in before.

src/iptvsimple/utilities/FileUtils.cpp
phunkyfish: Does this need to wait for filesystem changes to use the new STAT stuff?

src/iptvsimple/utilities/WebUtils.cpp
```diff
@@ -45,13 +44,13 @@ const std::string WebUtils::UrlEncode(const std::string& value)
...
   std::string strContent;
-  void* fileHandle = XBMC->OpenFile(url.c_str(), 0x08); //READ_NO_CACHE
-  if (fileHandle)
+  kodi::vfs::CFile file;
+
+  if (file.OpenFile(url, READ_NO_CACHE))
   {
     char buffer[1024];
-    if (int bytesRead = XBMC->ReadFile(fileHandle, buffer, 1024))
+    if (int bytesRead = file.Read(buffer, 1024))
```
phunkyfish, line 52: How come you don't close the file? Or does it auto close when destroyed?

phunkyfish:
> Is updated and test was OK, as part of filesystem it need to have the other request in before.

Cool

src/iptvsimple/utilities/WebUtils.cpp
AlwinEsch: It makes by destructor a auto close of file. Only sometimes I add .Close() it if the CFile use is on begin and bigger work behind his work or if CFile used for another file.

PR #367, opened 2020-04-27
Only allow timeshifting catchup from EPG if channel supports it

phunkyfish:
v4.13.4
- Fixed: Only allow timeshifting catchup from EPG if channel supports it

Use case missing from: https://github.com/kodi-pvr/pvr.iptvsimple/pull/346

### DIFF
--- a/pvr.iptvsimple/addon.xml.in
+++ b/pvr.iptvsimple/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.iptvsimple"
-  version="4.13.3"
+  version="4.13.4"
   name="PVR IPTV Simple Client"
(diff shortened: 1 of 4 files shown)


phunkyfish:
@AlwinEsch there is a client.cpp change in this PR so you will need to explicitly include in your C++ API PR.

phunkyfish:
@puenkitchen please test and confirm the fix, artifacts: https://jenkins.kodi.tv/blue/organizations/jenkins/kodi-pvr%2Fpvr.iptvsimple/detail/PR-367/1/artifacts

Please note that you must have **Play from EPG in live TV mode (using timeshift)** enabled to reproduce the use case.

puenktchen:
Well done. Your patch fixed the problem with "Play from EPG in live TV mode (using timeshift)" **enabled**
Thank you!

Issue #369, opened 2020-04-28
Matrix: Crash when trying to play stream from other Kodi addon


iranl:
IPTV Simple crashes Kodi (19.0-ALPHA1 Git:20200426-9216e28bc4, Windows 10 64-bit) when trying to play a video/stream from an other addon. 

The following works and plays correctly in Kodi 18.6 + IPTV Simple 3.9.8 but crashes in Kodi 19 Alpha + IPTV Simple 4.13.4. Example using plugin.video.youtube (but crash happens in all addons I tried).

To reproduce:
-Install Kodi 18.6 or Kodi 19 nightly 2020-04-26
-Install the latest available IPTV Simple for either version (3.9.8 or 4.13.4)
-Install the latest available plugin.video.youtube for either version (6.8.0)
-Set IPTV Simple to use the following playlist:

`#EXTM3U`
`#EXTINF:0 tvg-id="1" tvg-chno="1" tvg-name="Test Stream",Test Stream`
`plugin://plugin.video.youtube/play/?video_id=dQw4w9WgXcQ`

-Play the "Test Stream" channel through Kodi

[Debug.log](https://github.com/kodi-pvr/pvr.iptvsimple/files/4546983/kodi.log) -> Does not give much information at the time of the crash

phunkyfish:
Thanks for the really detailed report. Appears I broke this while adding inspection for stream URLs.

Fix: https://github.com/kodi-pvr/pvr.iptvsimple/pull/370

iranl:
Fixed in #370.

Closing.

PR #370, opened 2020-04-28
Fix plugins

phunkyfish:
v4.13.5
- Fixed: Fix plugin URLs for channels by disabling stream type URL tests
- Update: Update readme concerning terminating catchup streams


iranl:
Tested the fix for #369 using the Jenkins Artifact for Win x64. 

Working perfectly. Tnx for the quick fix

phunkyfish:
No problem, thank you for the clear bug report!

PR #387, opened 2020-05-16
[Matrix] API change update

AlwinEsch:
This changes are related to API change with xbmc/xbmc#17813

Need to bring in after them!

### DIFF
--- a/src/iptvsimple/utilities/FileUtils.h
+++ b/src/iptvsimple/utilities/FileUtils.h
@@ -12,6 +12,16 @@
 
 #include <string>
 
+// Prevent conflicts with Windows macros where have this names.
+#ifdef _WIN32 // windows
+#ifdef CreateDirectory
+#undef CreateDirectory
+#endif // CreateDirectory
+#ifdef DeleteFile
+#undef DeleteFile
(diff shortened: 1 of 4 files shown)


phunkyfish (approved):

src/iptvsimple/utilities/FileUtils.h
```diff
@@ -12,6 +12,15 @@
...
 #include <string>
 
+// Prevent conflicts with Windows macros where have this names.
+#ifdef CreateDirectory
+#undef CreateDirectory
+#endif // CreateDirectory
+#ifdef DeleteFile
+#undef DeleteFile
+#endif // DeleteFile
+#endif // _WIN32
```
MilhouseVH, line 23: http://ix.io/2mEz

AlwinEsch: Ups

AlwinEsch: Have added the missing `#ifdef`, this was moved as them to new headers and not seen by global jenkins test that there needed too, thats why now on addon.

Issue #393, opened 2020-05-29
Catchup: IPTV Manager handling


michaelarnauts:
## Background info

I'm trying to fit in the use case of IPTV Manager with the new catchup feature in Matrix. Note that I'm not using an IPTV provider, but I use locally installed Add-ons that are able to play legally available live streams, have access to EPG data and are able to play programs on demand.

See https://github.com/add-ons/service.iptv.manager for more information.

## Current implementation

So, I construct a m3u playlist with a list of channels, and an xmltv file with the epg data for these channels. Every channel has a `plugin://` url to play the channel live. For some programs, I have a `plugin://` url that I can use to play the program on-demand, but not for all programs. I can even have a `plugin://` url for programs in the future when they are already available (when it's a re-run for example).

I have added the catchup tags to configure catchup on the channel, and I use `catchup-id` as the full URL. This is not the intended usage for `catchup-id`, but it fits my purpose nicely, and seems to work fine.

Currently, I have added the following fields to a channel in the playlist:
* catchup="default"
* catchup-source="{catchup-id}" 
* catchup-days="10"

```
#EXTINF:-1 tvg-name="VTM" tvg-id="vtm.be" tvg-logo="special://home/addons/plugin.video.vtm.go/resources/logos/vtm.png" tvg-chno="3" group-title="VTM GO" catchup="default" catchup-source="{catchup-id}" catchup-days="10",VTM
plugin://plugin.video.vtm.go/play/catalog/channels/d8659669-b964-414c-aa9c-e31d8d15696b
```

Then, in the xmltv file, I have the following:

Currently, I have added the following fields to a channel in the playlist:
* catchup-id="plugin://plugin.video.example/play/12345678999"

```
<programme start="20200530202000 +0200" stop="20200530214500 +0200" channel="vtm.be" catchup-id="plugin://plugin.video.vtm.go/play/catalog/episodes/ff2f3192-d9b9-4ab3-80f3-70c2bf2d23fb">
 <title>Lego Masters</title>
 <desc>In deze finale is de opdracht: geen! Brickmaster geeft de finalisten maar liefst 28 uur om hun masterbuild te bouwen. Welk duo wint de 25.000 euro en de titel: LEGO Masters van Nederland en België?</desc>
[...]

PR #394, opened 2020-05-30
Use kodi's built in inputstreams for plugin:// urls

mediaminister:
This prevents that InputStream Adaptive is forced when using `plugin://` urls.


phunkyfish:
Nice, thanks for the PR.

Can you also add the changelog entry and update the addon xml version?

As soon as that’s updated correctly I can merge and release this.

michaelarnauts:
This fix is needed to play some channels with IPTV manager, although it could make sense for other usages also.

It might make sense to release this together with the changes requested in #393

phunkyfish:
If this can wait that’s ok with me.

phunkyfish (approved):

PR #395, opened 2020-06-01
Catchup fixups, plugin URLs and VOD catchup mode

phunkyfish:
v5.2.0
- Added: Enable timeshift feature from inputstream.ffmpegdirect
- Added: Support VOD catchup streams including those via plugin URLs
- Added: Access to ffmpegdirect settings and inputstream checks (C++ API)
- Fixed: Fix flussonic ts detected as wrong type
- Fixed: Use kodis built in inputstreams for plugin urls
- Fixed: Fix incorrect catchup granularity for flussonic stream

### DIFF
--- a/src/iptvsimple/data/Channel.cpp
+++ b/src/iptvsimple/data/Channel.cpp
@@ -211,6 +211,12 @@ bool Channel::IsCatchupSupported() const
   return Settings::GetInstance().IsCatchupEnabled() && m_hasCatchup && !m_catchupSource.empty();
 }
 
+bool Channel::SupportsLiveStreamTimeshifting() const
+{
+  return Settings::GetInstance().IsTimeshiftEnabled() && Settings::GetInstance().IsTimeshiftEnabledHttp() &&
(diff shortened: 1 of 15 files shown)


src/iptvsimple/data/Channel.cpp
```diff
@@ -198,10 +200,11 @@ void Channel::TryToAddPropertyAsHeader(const std::string& propertyName, const st
 
 void Channel::SetCatchupDays(int catchupDays)
 {
-  if (catchupDays != 0)
+  if (catchupDays > 0)
```
michaelarnauts: When `catchupDays` == -1 (ignore), this doesn't do anything.

phunkyfish: Faceplam

michaelarnauts: Hmm, this might be okay, but the default is set to 0 in Channel::Reset

phunkyfish: The default is fine. We will always set it for VOD type.

PR #401, opened 2020-06-30
Correctly pass realtime stream information

vpeter4:
discussion at https://github.com/xbmc/inputstream.ffmpegdirect/issues/57


phunkyfish:
Nice, can you update the version to 6.2.2 in `addon.xml.in` and the changelog entry to both that file and changelog.txt?

vpeter4:
PR updated.

phunkyfish:
Thanks for the contribution!

PR #408, opened 2020-07-22
Remove p8-platform dependency

basilgello:
This is a backport of AlwinEsch's commit in pvr.vuplus:

kodi-pvr/pvr.vuplus#8ff87ec16b7504522b97b40c40a50138f4f27ebc

Tested by building the Debian package with "kodiplatform" and "p8-platform"
build-dependencies removed from debian/control file.

### DIFF
--- a/src/iptvsimple/PlaylistLoader.cpp
+++ b/src/iptvsimple/PlaylistLoader.cpp
@@ -11,6 +11,7 @@
 #include "Settings.h"
 #include "utilities/FileUtils.h"
 #include "utilities/Logger.h"
+#include "utilities/StringUtils.h"
 #include "utilities/WebUtils.h"
 
 #include <chrono>
@@ -21,7 +22,6 @@
 #include <vector>
 
 #include <kodi/General.h>
-#include <p8-platform/util/StringUtils.h>
 
 using namespace iptvsimple;
 using namespace iptvsimple::data;
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,17 +4,14 @@ project(pvr.iptvsimple)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR})
 
 find_package(Kodi REQUIRED)
-find_package(p8-platform REQUIRED)
 find_package(Pugixml REQUIRED)
 find_package(ZLIB REQUIRED)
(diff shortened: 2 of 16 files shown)


src/iptvsimple/PlaylistLoader.cpp
```diff
@@ -21,7 +21,7 @@
 #include <vector>
 
 #include <kodi/General.h>
-#include <p8-platform/util/StringUtils.h>
+#include "utilities/StringUtils.h"
```
phunkyfish: As this is no longer a library the include needs to be moved to where the other source file headers are placed.

phunkyfish: This goes for all the files where this change was made.

phunkyfish (changes requested):

src/iptvsimple/PlaylistLoader.cpp
basilgello: Fixed, thanks! Principally, is this kind of change OK to be propagated across the rest of PVR add-ons using p8/kodi-platform?

basilgello:
I checked another PVR addons and some of them depend on multiple header files from p8. That means, orphaning and removing p8-platform package from Debian is unfeasible, in contra to kodiplatform which is small and embeddable. Since kodiplatform has already been removed from dependencies, there is nothing to do here so I close this PR.

@phunkyfish sorry for wasted time! Now hunting only for kodiplatform! :)

phunkyfish:
No problem. I think p8-platform will get removed eventually. But it can be a bit messy as it means moving the threading to C++11.

If you would still like to merge the changes you have made here I’m good with that as it needs to be done eventually anyway. So if you can reopen I can merge.

basilgello:
> No problem. I think p8-platform will get removed eventually. But it can be a bit messy as it means moving the threading to C++11.
> 
> If you would still like to merge the changes you have made here I’m good with that as it needs to be done eventually anyway. So if you can reopen I can merge.

Uhm, I can not reopen it after I removed the local repo :( Anyway, Ithink first we get rid of kodiplatform and then from p8. Removing p8 will not only require moving threads and os-specific helpers to c++11 but also fix libcec maintained by Pulse-Eight so I dont know if he'll agree to wipe it out...

phunkyfish:
No worries

PR #410, opened 2020-07-30
Allow max catchup of 15 days and premiere tag from XMLTV

phunkyfish:
v6.3.0
- Update: Allow max catchup of 15 days
- Added: Support premiere and new tag from XMLTV


phunkyfish:
@AlwinEsch there is something not working with the Azure pipelines. I don't know if it's random failures with dependencies or something else.

PR #412, opened 2020-08-02
Matrix [Fix Typo]

I-Cat:
Ok Looks Good.


I-Cat (commented):
yey

phunkyfish:
This PR needs to be a single commit. There is no reason to have 6 commits to change one typo.

Also, there is no point in reviewing your own PR. Someone else needs to review it that is not you.

I-Cat:
Okay. n.p off to the cmd I go

I-Cat:
@phunkyfish I Can't start a new pull request It says its locked?
thank you

PR #414, opened 2020-08-04
[Fix spelling and do proper pull request] File: Updated = Rytec-UK-Ireland.xml

I-Cat:
[Fix spelling and do proper pull request] File: Updated = Rytec-UK-Ireland.xml

yippy!
Don't know why It was giving me a hard time to fix this .

### DIFF
--- a/pvr.iptvsimple/resources/data/genres/genreTextMappings/Rytec-UK-Ireland.xml
+++ b/pvr.iptvsimple/resources/data/genres/genreTextMappings/Rytec-UK-Ireland.xml
@@ -9,7 +9,7 @@ Mapping Rytec Text Genres:
  - If a mapping cannot be found the text will be used but there will be no colouring (Genre Description only)
  - All text will be matched in lowercase
 
-If you are creating yoru own text mappings make a copy of this file in the same directory so it's not overwritten and start from there.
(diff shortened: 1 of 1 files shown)


I-Cat:
Mention @phunkyfish

AlwinEsch:
As note to git by previous request, there can be leaved on same request and branch. Already present code can be still on opened pull request changed.

E.g. about other request:
```
git reset --soft HEAD~6 # The reset with --soft not change the code only the related commits are unset, HEAD means uppest place and ~6 how many commits it should go back
git commit -m "Update Rytec-UK-Ireland.xml"
git push -f # The -f force to change existing code inside a branch
```

AlwinEsch:
To compare at home before a request is opened:
```
git remote add upstream https://github.com/kodi-pvr/pvr.iptvsimple
git fetch upstream
git format-patch upstream/Matrix # upstream to mark for other place and Matrix the wanted branch there
```
Name upstream can also another e.g. myotherplacename, is only a value name about URL

The `git format-patch` write then all commits to separate files and how they are then come on request.

phunkyfish (approved):
Thanks for the contribution

I-Cat:
@AlwinEsch 
@phunkyfish 

Thanks for helping!
😄

PR #430, opened 2020-09-16
[Author Change]

I-Cat:


phunkyfish:
Manual changes to other language files are not allowed. They are automated via transifix and only the en_GB should be modified.

PR #435, opened 2020-10-01
Cpp 17

phunkyfish:
v6.3.2
- Remove p8-platforms and add local StringUtils for cpp17

### DIFF
--- a/src/iptvsimple/utilities/StringUtils.cpp
+++ b/src/iptvsimple/utilities/StringUtils.cpp
@@ -1,0 +1,1149 @@
+/*
+ *      Copyright (C) 2005-2013 Team XBMC
+ *      http://xbmc.org
+ *
+ *  This Program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2, or (at your option)
+ *  any later version.
+ *
+ *  This Program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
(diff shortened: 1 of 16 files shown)


src/iptvsimple/utilities/StringUtils.cpp
```diff
@@ -0,0 +1,1149 @@
...
+// hack to ensure that std::string::iterator will be dereferenced as _unsigned_ char
+// without this hack "TrimX" functions failed on Win32 with UTF-8 strings
+static int isspace_c(char c)
+{
+  return ::isspace((unsigned char)c);
+}
+
+std::string& StringUtils::TrimLeft(std::string &str)
+{
+  str.erase(str.begin(), ::find_if(str.begin(), str.end(), [](char s) { return isspace_c(s) == 0; }));
```
AlwinEsch: You taken from latest p8-platform, this stupidly brings problems with older versions

Here how done inside Kodi itself https://github.com/xbmc/xbmc/blob/master/xbmc/utils/StringUtils.cpp#L478

fuzzard: Lambda works since C++11 standard, do you really need something older?

AlwinEsch: Was come to it with the build here https://travis-ci.org/github/kodi-pvr/pvr.hts/builds/730383117 where used it with this.

fuzzard: Fair enough then

phunkyfish: Actually because StringUtils is local now it doesn't require the older fix. The lambda works fine. So must be an issue with the p8-platform build for add-ons misinterpreting some build flags.

AlwinEsch (approved):

PR #449, opened 2020-11-15
 Set latest API stable inputstream version for Matrix

phunkyfish:
v7.1.0
- Added: Set minimum inputstream ffmpegdirect, rtmp and adaptive versions to API stable version for Matrix
- Added: Redact URLs when logged
- Update: Replace square brackets with bold text in addon settings help text


alexandrgoryachko:
@phunkyfish When will you release it?

phunkyfish:
Not until just before kodi beta2 will be released so it does not break any current beta1 users.

PR #456, opened 2020-12-06
Timestamp format

phunkyfish:
v7.2.0
- Added: Safe localtime and format strings for timestamps in catchup
- Added: Allow now/current time format specifiers in live stream URLs
- Added: Redact catchup source and url format string
- Added: Some providers use 'tvg-rec' instead of 'catchup-days'
- Added: Set min ffmpegdirect version to 1.19.0 so it supports timestamp format


bkuhls:
Not a bug report, just an info: This PR breaks building with gcc-4.9:
```
src/iptvsimple/CatchupController.cpp:292:11: error:
 'put_time' is not a member of 'std'
```
because std::put_time was added with gcc-5.0:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=54354#c9

phunkyfish:
put_time is part of the c++11 spec. Is gcc 4.9 not a full implementation?

phunkyfish:
It appears you need gcc5

bkuhls:
> It appears you need gcc5

Right, or something like this: https://github.com/bkuhls/buildroot/blob/9b2509f58e013168b9dfc59bff2623b64c84f035/package/kodi-pvr-iptvsimple/0001-CatchupController-add-put_time-to-support-gcc-5.patch

phunkyfish:
Which OS are you using. Ubuntu 16.04 has 5.4 available I thought?

bkuhls:
> Which OS are you using. Ubuntu 16.04 has 5.4 available I thought?

I am maintaining the kodi package for the buildroot project which aims at cross-compiling for embedded platforms. One of my target devices, although running kernel 5.10, is stuck at gcc-4.9...

phunkyfish:
Gotcha. So you can include the above patch for your project?

bkuhls:
> Gotcha. So you can include the above patch for your project?

yes, build-test was positive, run-time test is pending, also for inputstream.ffmpegdirect.

Just wanted to get tell about the problem so others can use the "fix" I found.
Maybe the project can use something like this: https://lab.louiz.org/louiz/biboumi/commit/a195939cdaca343013d32ae902839ffdb28c33cb or add the put_time definition to some header in include/kodi/tools/ to be used by all kodi addons. But I do not expect it ;) due to Kodi moving to c++17: https://forum.kodi.tv/showthread.php?tid=358086&pid=2987011#pid2987011

phunkyfish:
You’re right there. We need to keep to the standard and instep with kodi due to the binary nature of the add-ons. So once kodi goes to Cpp17 so will the binary add-ons. This will happen first thing in kodi 20.

I moved away from gmtime and strftime implementations due to erroneous cross platform bugs, mainly on windows which are very hard to find and debug. And nevermind the thread safety issue which once more are hard to find/debug.

PR #457, opened 2021-01-02
some providers use 'tvg-rec' instead of 'catchup-days'

arthur-liberman:
I just found that some providers use `tvg-rec` to indicate the timeshift buffer in their archive functionality, instead of `catchup-days`
This PR uses `tvg-rec` as a fallback if `catchup-days` is not available.

### DIFF
--- a/src/iptvsimple/PlaylistLoader.cpp
+++ b/src/iptvsimple/PlaylistLoader.cpp
@@ -196,6 +196,7 @@ std::string PlaylistLoader::ParseIntoChannel(const std::string& line, Channel& c
     std::string strTvgShift   = ReadMarkerValue(infoLine, TVG_INFO_SHIFT_MARKER);
     std::string strCatchup       = ReadMarkerValue(infoLine, CATCHUP);
     std::string strCatchupDays   = ReadMarkerValue(infoLine, CATCHUP_DAYS);
+    std::string strTvgRec        = ReadMarkerValue(infoLine, TVG_INFO_REC);
     std::string strCatchupSource = ReadMarkerValue(infoLine, CATCHUP_SOURCE);
     std::string strCatchupSiptv = ReadMarkerValue(infoLine, CATCHUP_SIPTV);
     std::string strCatchupCorrection = ReadMarkerValue(infoLine, CATCHUP_CORRECTION);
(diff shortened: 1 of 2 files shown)


src/iptvsimple/PlaylistLoader.h
```diff
@@ -27,6 +27,7 @@ namespace iptvsimple
   static const std::string TVG_INFO_LOGO_MARKER    = "tvg-logo=";
   static const std::string TVG_INFO_SHIFT_MARKER   = "tvg-shift=";
   static const std::string TVG_INFO_CHNO_MARKER    = "tvg-chno=";
+  static const std::string TVG_INFO_REC            = "tvg-rec="; //some provider use 'tvg-rec' instead of 'catchup-days'
```
phunkyfish: `// some providers use 'tvg-rec' instead of 'catchup-days'`

src/iptvsimple/PlaylistLoader.cpp
```diff
@@ -196,6 +196,8 @@ std::string PlaylistLoader::ParseIntoChannel(const std::string& line, Channel& c
     std::string strTvgShift   = ReadMarkerValue(infoLine, TVG_INFO_SHIFT_MARKER);
     std::string strCatchup       = ReadMarkerValue(infoLine, CATCHUP);
     std::string strCatchupDays   = ReadMarkerValue(infoLine, CATCHUP_DAYS);
+    if (strCatchupDays.empty())
+      strCatchupDays = ReadMarkerValue(infoLine, TVG_INFO_REC);
```
phunkyfish: Please just read the value into its own variable here and then add any logic here: https://github.com/kodi-pvr/pvr.iptvsimple/blob/98302028dd54789b3b0f2d134f09d49284e9c8e4/src/iptvsimple/PlaylistLoader.cpp#L259-L269

Otherwise the logic around catchup days will be scattered. Better to have it in one place.

phunkyfish (changes requested):

src/iptvsimple/PlaylistLoader.h
arthur-liberman: Copy paste is bad, roger ;)

arthur-liberman:
Both change requests pushed

phunkyfish (approved):
Thanks for contribution. No need to add addon xml version etc here as I have another PR to go in so can just merge this before that in a few days time.

PR #458, opened 2021-01-03
Add support for groups defined by the #EXTGRP tag

arthur-liberman:


arthur-liberman:
I'll reopen the PR after I fix it.

PR #469, opened 2021-01-24
Epg load speed

phunkyfish:
v7.4.0
- Fixed: An xmltv channel cannot have an empty id field
- Fixed: Fix slow epg load due to display name checks when finding a channel
- Fixed: Only force load EPG data on startup if catchup is enabled
- Added: Support custom channel group filters
- Added: Reload settings before reloading playlist and EPG data


bantilan (approved):
working on Rockchip Tinker board and RPI 3 B+

phunkyfish:
Thanks for testing!

phunkyfish:
I have also added a new feature which allows specifying a custom list of groups names. Both in the UI and via a config file if many are required.

PR #485, opened 2021-02-22
Fix terminating catchup

phunkyfish:
v7.4.3
- Fixed: Add support for format specifiers use for detecting terminating catchup and granularity
- Update: Make inputstream add-ons a required dependency


phunkyfish:
@batelcho please try this. 

You can find the testbuilds here: https://jenkins.kodi.tv/blue/organizations/jenkins/kodi-pvr%2Fpvr.iptvsimple/detail/PR-485/1/artifacts

PR #494, opened 2021-03-13
[lang] Update language for zh_cn

Meano:
Update zh_cn language file for pvr.iptvsimple.

### DIFF
--- a/pvr.iptvsimple/resources/language/resource.language.zh_cn/strings.po
+++ b/pvr.iptvsimple/resources/language/resource.language.zh_cn/strings.po
@@ -16,58 +16,408 @@ msgstr ""
 "Language: zh_CN\n"
 "Plural-Forms: nplurals=1; plural=0;\n"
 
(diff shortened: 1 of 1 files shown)


phunkyfish:
Only the en_gb file can be update here. All other languages are auto translated by transifix.

Meano:
> Only the en_gb file can be update here. All other languages are auto translated by transifix.

Oh, there's no access permission in transifex KODI Main project, and nobody reply the language request...

phunkyfish:
Let me ask.

phunkyfish:
We are migrating to a new translation tool. So unless someone picks up the language request it will have to wait until the migration is completed. Check back on this thread in a few weeks and I’ll see what progress we have.

Meano:
OK, thanks a lot!

Meano:
I'm updating the translation on weblate, this PR is not needed anymore.

PR #523, opened 2021-06-08
Update chinese translate strings for iptv plugin

SummonHIM:


phunkyfish:
Only en_gb are updated in the repo. All others are translated by transifix. Soon moving to weblate.

SummonHIM:
oh sure

PR #549, opened 2021-08-29
Return server error if channels or groups could not be loaded due to …

popy2k14:
…missing file so they are not cleared in Kodi

@phunkyfish 
Is an jenkins build possible to test it?
Sadly dont have an android dev environment.

Also, can you please do a review?

Thanks a lot


phunkyfish:
Just click Jenkins under the checks and then artifacts to download the test builds. It all runs as part of the github PR process.

Screenshot 2021-08-29 at 20 47 48

PR #552, opened 2021-09-10
Fix <date> parsing only year to firstAired

matthuisman:
wrong format template causing only the year to be picked up by kodi

This will also be stopping any epg entries being marked as "new"
as this line currently will always be false: https://github.com/kodi-pvr/pvr.iptvsimple/blob/Matrix/src/iptvsimple/data/EpgEntry.cpp#L230

Before:
screenshot00003
After:
screenshot00002

(need to PVR > Clear data after updating)

### DIFF
--- a/pvr.iptvsimple/addon.xml.in
+++ b/pvr.iptvsimple/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.iptvsimple"
-  version="7.6.10"
+  version="7.6.11"
   name="PVR IPTV Simple Client"
   provider-name="nightik and Ross Nicholson">
   <requires>@ADDON_DEPENDS@
@@ -21,6 +21,9 @@
       <icon>icon.png</icon>
     </assets>
     <news>
+v7.6.11
+- Fixed: EPG date entry only parsing year
+
 v7.6.10
 - Fixed: Fix episode number when there is no season
 - Fixed: Return server error if channels or groups could not be loaded due to missing file so they are not cleared in Kodi
--- a/pvr.iptvsimple/changelog.txt
+++ b/pvr.iptvsimple/changelog.txt
@@ -1,3 +1,6 @@
+v7.6.11
+- Fixed: EPG date entry only parsing year
+
 v7.6.10
 - Fixed: Fix episode number when there is no season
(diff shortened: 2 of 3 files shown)


phunkyfish:
Looks good, just update the changelog and addon xml and it's ready.

matthuisman:
done.
just doing a quick run-time test.

Also, keep in mind when it branches to Nexus,
all string:formats can be changed to the new format
https://github.com/xbmc/xbmc/commit/100c53a30a4de163cad44fee9a802df85bd8feb7#diff-4c4adde65e7d469674f53e049fbb67b2148b1bd97a82e2f193a74447c368dd46L88
In this case, I assume it would be "{:04d}-{:02d}-{:02d}"

matthuisman:
@phunkyfish 
ready to go :)

found another sort of bug

date needs to be in users local time format (not possible when using external xmls)
otherwise it won't match the start date if thats in utc for example.
When simple compares them, its already converted the <start> into local
and just compares the start formated to yyyy-mm-dd with the date converted to yyyy-mm-dd
However, date is timezone unaware..
So, really it should more or less compare the "raw" start yyyy-mm-dd with the date

phunkyfish:
> done.
> just doing a quick run-time test.
> 
> Also, keep in mind when it branches to Nexus,
> all string:formats can be changed to the new format
> [xbmc/xbmc@100c53a#diff-4c4adde65e7d469674f53e049fbb67b2148b1bd97a82e2f193a74447c368dd46L88](https://github.com/xbmc/xbmc/commit/100c53a30a4de163cad44fee9a802df85bd8feb7#diff-4c4adde65e7d469674f53e049fbb67b2148b1bd97a82e2f193a74447c368dd46L88)
> In this case, I assume it would be "{:04d}-{:02d}-{:02d}"

I’m not sure on this one. Binary addons are special. But easy to try once on nexus.

phunkyfish:
> @phunkyfish
> ready to go :)
> 
> found another sort of bug
> 
> date needs to be in users local time format (not possible when using external xmls)
> otherwise it won't match the start date if thats in utc for example.
> When simple compares them, its already converted the into local
> and just compares the start formated to yyyy-mm-dd with the date converted to yyyy-mm-dd
> However, date is timezone unaware..
> So, really it should more or less compare the "raw" start yyyy-mm-dd with the date

Maybe create an issue to discuss this one or ping me on slack.

phunkyfish (approved):
LGTM

PR #557, opened 2021-09-11
use tz aware <date> for m_firstAired

### LINKED ISSUE #553: Issue with new tag comparing <date> with start time
image

```
<programme start="20210910180000 +0000" stop="20210910190000 +0000">
<date>20210910</date>
```

when simple is comparing the start time with the <date - the start has been converted to local time.
https://github.com/kodi-pvr/pvr.iptvsimple/blob/Matrix/src/iptvsimple/data/EpgEntry.cpp#L230

So, in the above and my timezone, it's now 2021-09-11

However, date is timezone unaware. so often the match won't work.

Simple should be comparing the "raw" start yyyy-mm-dd with the date instead.
That way they are both treated in the same timezone

So, grab the yyyy-mm-dd from 20210910180000 +0000
and then compare that with the yyyy-mm-dd from date

matthuisman:
Fixes: https://github.com/kodi-pvr/pvr.iptvsimple/issues/553

Before
before
After
after

### DIFF
--- a/src/iptvsimple/data/EpgEntry.cpp
+++ b/src/iptvsimple/data/EpgEntry.cpp
@@ -226,10 +226,18 @@ bool EpgEntry::UpdateFrom(const xml_node& channelNode, const std::string& id,
     static const std::regex dateRegex("^[1-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9]");
     if (std::regex_match(dateString, dateRegex))
     {
-      m_firstAired = ParseAsW3CDateString(dateString);
-      // Always compare to the raw string date value to avoid timezone issues
-      if (m_firstAired == ParseAsW3CDateString(strStart))
-        m_new = true;
+      long long tmpDate = ParseDateTime(dateString.substr(0,8) + strStart.substr(8));
+      // Windows localtime_s does not support negative time_t
+      if (tmpDate < 0)
+      {
+        m_firstAired = ParseAsW3CDateString(dateString);
+        m_new = m_firstAired == ParseAsW3CDateString(strStart);
+      }
+      else
+      {
+        m_firstAired = ParseAsW3CDateString(static_cast<time_t>(tmpDate));
+        m_new = m_firstAired == ParseAsW3CDateString(m_startTime);
+      }
     }
 
     std::sscanf(dateString.c_str(), "%04d", &m_year);


src/iptvsimple/data/EpgEntry.cpp
```diff
@@ -220,15 +220,20 @@ bool EpgEntry::UpdateFrom(const xml_node& channelNode, const std::string& id,
...
-      m_firstAired = ParseAsW3CDateString(dateString);
-      // Always compare to the raw string date value to avoid timezone issues
-      if (m_firstAired == ParseAsW3CDateString(strStart))
+      dateString = dateString.substr(0,8) + strStart.substr(8);
+      long long tmpDate = ParseDateTime(dateString);
+      // Windows localtime_s doesnt support negative time_t
+      if tmpDate < 0
+        m_firstAired = ParseAsW3CDateString(dateString);
+      else
+        m_firstAired = ParseAsW3CDateString(static_cast<time_t>(tmpDate));
```
phunkyfish, line 238: You could just embed the ParseDateTime(dateString) call here, no need for tempDate var

src/iptvsimple/data/EpgEntry.cpp
```diff
@@ -220,15 +220,20 @@ bool EpgEntry::UpdateFrom(const xml_node& channelNode, const std::string& id,
...
-      // Always compare to the raw string date value to avoid timezone issues
-      if (m_firstAired == ParseAsW3CDateString(strStart))
+      dateString = dateString.substr(0,8) + strStart.substr(8);
+      long long tmpDate = ParseDateTime(dateString);
+      // Windows localtime_s doesnt support negative time_t
+      if (tmpDate < 0)
+        m_firstAired = ParseAsW3CDateString(dateString);
+      else
+        m_firstAired = ParseAsW3CDateString(static_cast<time_t>(tmpDate));
+      if (m_firstAired == ParseAsW3CDateString(m_startTime))
```
phunkyfish: You will still want to compare this is strStart right?

src/iptvsimple/data/EpgEntry.cpp
matthuisman: we need tmpDate to check if its before epoch :) (the if statement above it)

src/iptvsimple/data/EpgEntry.cpp
phunkyfish: Or actually the check needs to depend on which branch it takes for m_firstAired

matthuisman: nope, as that wouldn't be in "local time". 
We want to now compare both local times
only time it won't work is for dates before 1970
we wont have m_startTime going back that far anyway

matthuisman: if you wanted it 100% even showing "new" for epg entries from 1960, you would do
```
      if (tmpDate < 0) {
        m_firstAired = ParseAsW3CDateString(dateString);
        if (m_firstAired == ParseAsW3CDateString(strStart))
            m_new = true;
        }
      else{
        m_firstAired = ParseAsW3CDateString(static_cast<time_t>(tmpDate));
        if (m_firstAired == ParseAsW3CDateString(m_startTime))
            m_new = true;
        }
```

src/iptvsimple/data/EpgEntry.cpp
phunkyfish: Opps, so we do.

src/iptvsimple/data/EpgEntry.cpp
phunkyfish: Yes, this is what I meant. But it could be shorter:

```
      if (tmpDate < 0) 
      {
        m_firstAired = ParseAsW3CDateString(dateString);
        m_new = m_firstAired == ParseAsW3CDateString(strStart);
      }
      else
      {
        m_firstAired = ParseAsW3CDateString(static_cast<time_t>(tmpDate));
        m_new = m_firstAired == ParseAsW3CDateString(m_startTime);
      }
```

src/iptvsimple/data/EpgEntry.cpp
```diff
@@ -220,15 +220,20 @@ bool EpgEntry::UpdateFrom(const xml_node& channelNode, const std::string& id,
...
+  std::string dateString = GetNodeValue(channelNode, "date");
   if (!dateString.empty())
   {
     static const std::regex dateRegex("^[1-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9]");
     if (std::regex_match(dateString, dateRegex))
     {
-      m_firstAired = ParseAsW3CDateString(dateString);
-      // Always compare to the raw string date value to avoid timezone issues
-      if (m_firstAired == ParseAsW3CDateString(strStart))
+      dateString = dateString.substr(0,8) + strStart.substr(8);
```
phunkyfish: This should be a new string, maybe called dateTimeString and should be const.

src/iptvsimple/data/EpgEntry.cpp
```diff
@@ -220,15 +220,20 @@ bool EpgEntry::UpdateFrom(const xml_node& channelNode, const std::string& id,
 
   m_genreString = GetJoinedNodeValues(channelNode, "category");
 
-  const std::string dateString = GetNodeValue(channelNode, "date");
+  std::string dateString = GetNodeValue(channelNode, "date");
```
phunkyfish: Leave this as const

src/iptvsimple/data/EpgEntry.cpp
matthuisman: oooo, thats pretty :)
done!
(even though a m_firstAired of before 1970 would be odd!)

src/iptvsimple/data/EpgEntry.cpp
matthuisman: done :)

phunkyfish (approved):

PR #561, opened 2021-09-18
[Matrix] final Matrix change to correct test builds and take as Version 19.0.0

AlwinEsch:
This change the builds to final released Kodi Matrix.

Further is the version to 19.0.0 increased to have equal to Kodi and to see on which Version this addon works.


phunkyfish (approved):
👍

PR #609, opened 2022-06-18
Fix "ch-number" tag being ignored

vasilevp:


phunkyfish:
Thanks for the fix. Before this can be merged the fix needs to be done for Nexus branch. In addition the `addon.xml.in` requires the micro number to be incremented and a entry added to `changelog.txt`.

phunkyfish:
I had some spare time today so I've created the PR for Nexus. I'll just merge and release this PR while I'm at it. Should appear as an update in the next 24 hours.

phunkyfish (approved):

PR #631, opened 2022-08-10
Support url arguments when encoding file portion of icon URL - Matrix

phunkyfish:
v19.2.1
- Support url arguments when encoding file portion of icon URL


phunkyfish:
Azure builds failing due to old image being deprecated.

PR #677, opened 2022-12-31
Ensure groups are delimited by semi-colon's when supplied by an EXTGRP tag

phunkyfish:
v20.6.1
- Ensure groups are delimited by semi-colon's when suoplied by an EXTGRP tag
- Fix warning from using sprintf by switching to snprintf


egorenar:
Thank you! Tested on x86_64 Linux.
Fixes my use case where a channel group specified by #EXTGRP contains spaces.

PR #693, opened 2023-02-05
Add setting to make matching channel IDs and tvg-id's case sensitive

phunkyfish:
v20.7.0
- Add setting to make matching channel IDs and tvg-id's case sensitive


alanmilinovic:
Tested and can be merged.

PR #696, opened 2023-02-07
Update strings.po

fanmingming:
The translation of some simplified Chinese parts has been improved.


gade01:
Please use Weblate for translations. 

https://kodi.weblate.cloud/projects/kodi-add-ons-pvr-clients/pvr-iptvsimple/zh_cn/

Note that the component has not yet been moved to Nexus branch, but will within the next week or so. 
This can cause some translations to appear missing, if they are not in sync with Matrix branch.

PR #802, opened 2023-11-06
Fix azure and gihub build runners on Nexus

phunkyfish:


ksooo:
@phunkyfish let me handle Nexus for all add-ons, please.

PR #818, opened 2023-12-18
Update FR strings.po

ludovik35:
Global update with all missing items translation


phunkyfish:
Updates for languages other than en_GB need to be made using weblate.

follow the instructions here to sign and contribute: https://kodi.tv/article/switching-translations-to-weblate/

PR #823, opened 2024-01-01
Add connection manager support to wait for a valid M3U file before starting the add-on instance

phunkyfish:
v21.7.0
- Add connection manager support to wait for a valid M3U file before starting the add-on instance
- The minimum refresh interval for the M3U should be 1 minutes and not zero, as that would be infinite refresh

### DIFF
--- a/pvr.iptvsimple/resources/instance-settings.xml
+++ b/pvr.iptvsimple/resources/instance-settings.xml
@@ -77,7 +77,7 @@
           <level>1</level>
           <default>60</default>
           <constraints>
-            <minimum>0</minimum>
+            <minimum>1</minimum>
             <step>10</step>
             <maximum>120</maximum>
           </constraints>
--- a/pvr.iptvsimple/resources/instance-settings.xml
+++ b/pvr.iptvsimple/resources/instance-settings.xml
@@ -106,7 +106,36 @@
         </setting>
       </group>
 
-      <group id="3" label="30006">
+      <group id="3" label="30078">
+        <setting id="connectioncheckinterval" type="integer" label="30080" help="30629">
+          <level>1</level>
+          <default>10</default>
+          <constraints>
+            <minimum>1</minimum>
+            <step>1</step>
+            <maximum>60</maximum>
+          </constraints>
+          <control type="slider" format="integer">
+            <popup>true</popup>
+            <formatlabel>14045</formatlabel>
+          </control>
+        </setting>
+        <setting id="connectionchecktimeout" type="integer" label="30079" help="30628">
+          <level>2</level>
+          <default>20</default>
+          <constraints>
+            <minimum>1</minimum>
+            <step>1</step>
+            <maximum>60</maximum>
+          </constraints>
+          <control type="slider" format="integer">
+            <popup>true</popup>
+            <formatlabel>14045</formatlabel>
+          </control>
+        </setting>
+      </group>
+
+      <group id="4" label="30006">
         <setting id="defaultProviderName" type="string" label="30007" help="30740">
           <level>2</level>
           <default></default>
--- a/pvr.iptvsimple/resources/language/resource.language.en_gb/strings.po
+++ b/pvr.iptvsimple/resources/language/resource.language.en_gb/strings.po
@@ -434,7 +434,22 @@ msgctxt "#30077"
 msgid "Ignore Case for EPG Channel IDs"
 msgstr ""
 
-#empty strings from id 30078 to 30099
+#. label-group: General - M3U Startup Check
+msgctxt "#30078"
+msgid "M3U Startup Check"
+msgstr ""
+
+#. label: Catchup - connectCheckTimoutSecs
+msgctxt "#30079"
+msgid "Timeout for check"
+msgstr ""
+
+#. label: Catchup - connectCheckIntervalSecs
+msgctxt "#30080"
+msgid "Interval for check"
+msgstr ""
+
+#empty strings from id 30081 to 30099
 
 #. label-category: catchup
 #. label-group: Catchup - Catchup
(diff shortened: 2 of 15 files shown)


pvr.iptvsimple/resources/language/resource.language.en_gb/strings.po
```diff
@@ -828,7 +843,17 @@ msgctxt "#30627"
...
-#empty strings from id 30628 to 30639
+#. help: EPG Settings - connectionCheckTimeoutSecs
+msgctxt "#30628"
+msgid "When checking for a valid M3U file, the length of time to wait before timing out the attempt."
+msgstr ""
+
+#. help: EPG Settings - connectionCheckIntervalSecs
+msgctxt "#30629"
+msgid "When checking for a valid M3U file, the length of time to wait between attempts. Note that a valid file will only be checked for on startup and once a valid file is foudn all checks stop."
+msgstr ""
```
matthuisman, line 854: small typo. "foudn" != "found"

src/iptvsimple/ConnectionManager.cpp
```diff
@@ -0,0 +1,181 @@
...
+    {
+      m_connectionListener.ConnectionLost();
+    }
+    else if (newState == PVR_CONNECTION_STATE_CONNECTED)
+    {
+      m_connectionListener.ConnectionEstablished();
+    }
+
+    /* Notify connection state change (callback!) */
+    m_connectionListener.ConnectionStateChange(m_settings->GetM3ULocation(), newState, "");
```
matthuisman: the notification may be slightly annoying? Do you think its required in this instance?

pvr.iptvsimple/resources/instance-settings.xml
```diff
@@ -106,7 +106,36 @@
...
+          <control type="slider" format="integer">
+            <popup>true</popup>
+            <formatlabel>14045</formatlabel>
+          </control>
+        </setting>
+        <setting id="connectionchecktimeout" type="integer" label="30079" help="30628">
+          <level>2</level>
+          <default>20</default>
+          <constraints>
+            <minimum>10</minimum>
```
matthuisman: can we please allow this to go down to 1 ?

matthuisman: also, the m3uRefreshIntervalMins allows going to 0 which i believe may be a mistake? 
That just makes it constantly refresh.
Maybe refresh interval should have a minimum of 1 or even 10

pvr.iptvsimple/resources/language/resource.language.en_gb/strings.po
phunkyfish: Fixed, ty

pvr.iptvsimple/resources/instance-settings.xml
phunkyfish: So this is essentially a TCP timeout. In the majority of cases it will just return immediately as you will have seen. The minimum is 10 so any TCP stuff can finish elegantly, lower values than that can be messy generally.

It's really the interval that we care about here.

phunkyfish: Point taken on the refresh, that would be once a second which is far too often. Will change the minimum there to 1.

src/iptvsimple/ConnectionManager.cpp
phunkyfish: Ok, moved this behind a variable for now. We may want this in the future if we extend functionality.

phunkyfish:
New testbuilds: https://jenkins.kodi.tv/blue/organizations/jenkins/kodi-pvr%2Fpvr.iptvsimple/detail/PR-823/3/artifacts

Question, what behaviour do you get/expect if there is no M3U URL set?

pvr.iptvsimple/resources/instance-settings.xml
matthuisman: ive found with a local server, if its just a second too early, itll only timeout after 10s (if port not available)
For a local server - 1s is huge amount.
Is there any harm letting it go lower?

phunkyfish: True, I guess there is no harm. Try and let me know how it works. I'll have to change the step to 1 also.

phunkyfish:
> Question, what behaviour do you get/expect if there is no M3U URL set?

Any response here? Actually NVM. I'll just have it wait for URL to be set before validating it.

New testbuilds: https://jenkins.kodi.tv/blue/organizations/jenkins/kodi-pvr%2Fpvr.iptvsimple/detail/PR-823/5/artifacts

matthuisman:
hmm. something seems off now. after booting - i get empty channels and epg.
I see it being loaded in the log though.
If i go Clear PVR data then it all appears

It doesnt show as pvr manager starting up either like it used too.

I went back to https://jenkins.kodi.tv/blue/organizations/jenkins/kodi-pvr%2Fpvr.iptvsimple/detail/PR-823/1/artifacts and that works

https://jenkins.kodi.tv/blue/organizations/jenkins/kodi-pvr%2Fpvr.iptvsimple/detail/PR-823/2 has no windows x64
but /3 is were it "breaks"

phunkyfish:
Hmmm, I won’t have history so hard to figure out the change that breaks it.

Can you explain your set up to me?

src/iptvsimple/ConnectionManager.cpp
phunkyfish: Reverted this change as the state change is required.

phunkyfish:
I bet it's because we removed the notification. My guess is it also updates Kodi PVR of the state change.

Yup, that was it. New builds: https://jenkins.kodi.tv/blue/organizations/jenkins/kodi-pvr%2Fpvr.iptvsimple/detail/PR-823/8/artifacts

matthuisman:
thank you. have tested and works good now :)

matthuisman:
@phunkyfish I forgot to test local paths! 

Just tried them now and they dont seem to work.

Localpath: special://userdata/addon_data/plugin.program.iptv.merge/playlist.m3u8

Just constantly get logged:
AddOnLog: pvr.iptvsimple: pvr.iptvsimple - Check Unable to open url: special://userdata/addon_data/plugin.program.iptv.merge/playlist.m3u8

I still think be nice if it retried for local paths as well, but maybe the check method needs to handle local paths?

phunkyfish:
I implemented it using local paths, but just the regular filesystem and not special paths. They should just work too though.

Will take a look later.

phunkyfish:
Please test the fix here: https://github.com/kodi-pvr/pvr.iptvsimple/pull/826

PR #828, opened 2024-01-28
Test EPG as Live build

phunkyfish:


phunkyfish:
Superseded by https://github.com/kodi-pvr/pvr.iptvsimple/pull/831

PR #845, opened 2024-03-06
Findlzma add alternate name to satisfy windows builds

fuzzard:
Windows builds that have lzma built before pvr.iptvsimple dont build xz-utils, and therefore the find_package is reliant on finding out of tree. The search name requires liblzma for additional assistance finding the lzma lib.

Locally this fixes the UWP build issue with all addons being built
Jenkins job at https://jenkins.kodi.tv/job/WIN-UWP-64/25024/ building inputstream.ffmpegdirect and pvr.iptvsimple (minimal repro case for any windows arch build)

### DIFF
--- a/Findlzma.cmake
+++ b/Findlzma.cmake
@@ -6,7 +6,7 @@
 # LZMA_LIBRARIES - the xz library
 
 find_path(LZMA_INCLUDE_DIRS lzma.h)
-find_library(LZMA_LIBRARIES lzma)
+find_library(LZMA_LIBRARIES NAMES lzma liblzma)
 
 include(FindPackageHandleStandardArgs)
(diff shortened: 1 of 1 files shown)


fuzzard:
test build is available on mirrors at https://mirrors.kodi.tv/test-builds/windows/uwp64/Kodi-20240306-1cd0b6bc-temp_repo_binary_addon-fuzzard-x64.msix
both addons are correctly in the msix

I'll also add, a baseline test with jenkins changes to allow us to more accurately target some addons for windows (rather than all, always) - https://jenkins.kodi.tv/job/WIN-UWP-64/25023/
that shows the current failure on master

phunkyfish:
Excellent, this will likely allow a lot of cleanup of the custom cmake for xz-utils for windows too.

phunkyfish (approved):

PR #849, opened 2024-03-30
Make LZMA support optional

heitbaum:
With the malicious actor changes to upstream xz, make the LZMA library optional. .xz support in pvr.iptvsimple is limited to epg files.
- cross-ref: https://github.com/LibreELEC/LibreELEC.tv/pull/8775

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,20 +6,17 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR})
 find_package(Kodi REQUIRED)
 find_package(pugixml REQUIRED)
 find_package(ZLIB REQUIRED)
-find_package(lzma REQUIRED)
+find_package(lzma)
 
 include_directories(${KODI_INCLUDE_DIR}/.. # Hack way with "/..", need bigger Kodi cmake rework to match right include ways
                     ${PUGIXML_INCLUDE_DIRS}
-                    ${ZLIB_INCLUDE_DIRS}
-                    ${LZMA_INCLUDE_DIRS})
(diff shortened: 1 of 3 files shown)


AlwinEsch:
**NOTE:**: There also xz-utils 5.4.3 where use the bad checksum and possible download from dangerous source place! Should changed too.
```
-- verifying file...
       file='C:/jenkins-workspace/workspace/binary-addons/kodi-windows-i686-Omega/cmake/addons/build/download/xz-5.4.3.tar.xz'
-- SHA256 hash of
    C:/jenkins-workspace/workspace/binary-addons/kodi-windows-i686-Omega/cmake/addons/build/download/xz-5.4.3.tar.xz
  does not match expected value
    expected: '92177bef62c3824b4badc524f8abcce54a20b7dbcfb84cde0a2eb8b49159518c'
      actual: '6732bd6c5e2cc7f520a645bdb006a286c7790f73950213cd59d2a9e4e4fd3d36'
```

AlwinEsch:
> **NOTE:**: There also xz-utils 5.4.3 where use the bad checksum and possible download from dangerous source place! Should changed too.
>....

@phunkyfish created a request here about https://github.com/kodi-pvr/pvr.iptvsimple/pull/850

phunkyfish:
> > **NOTE:**: There also xz-utils 5.4.3 where use the bad checksum and possible download from dangerous source place! Should changed too.
> > ....
> 
> @phunkyfish created a request here about #850

This won’t work unfortunately. 5.4.1 won’t build for win UWP. So we’ll just need to upload 5.4.3 to our kodi mirrors

phunkyfish:
I think is too much of a change, we don't link xz with sshd so this is a non-issue IMHO. Plus we are using an older version which should be safe anyway.

PR #850, opened 2024-03-30
Update xzutils to match kodi mirrors

phunkyfish:
v21.8.3
- Change xz-utils to 5.4.3 to match kodi mirrors


phunkyfish:
@AlwinEsch this appears to break Win32 and Win64 UWP. So might need to upload 5.4.3 to our mirrors.

AlwinEsch:
> @AlwinEsch this appears to break Win32 and Win64 UWP. So might need to upload 5.4.3 to our mirrors.

Puhh, oh yes the UWP, then we need to become it there.

PR #851, opened 2024-04-12
[StreamUtils] Removed inputstream.adaptive.manifest_update_parameter

CastagnaIT:
The "full" value support has been removed from InputStream Adaptive for Kodi 21, because managed automatically

ref. https://github.com/xbmc/inputstream.adaptive/wiki/Integration#inputstreamadaptivemanifest_update_parameter

Does not cause any problem but better to remove it,
does not require immediate binary bump, you can do it at the first opportunity


phunkyfish (approved):

phunkyfish:
Thanks!

PR #884, opened 2024-08-22
Update reading XMLTV xmltv_ns for PVR API v9.0.0

emveepee:
Update EpisodePartNumber with the total number of episodes in an episode sequence.  Use with Episode number

As an example from this XMLTV programme the programs shows E45/48 in the Estuary skin

```
	<programme start="20240823161500 +0000" stop="20240823170000 +0000" channel="nextpvr-7400">
		<title>Pointless</title>
		<desc>45/48. Alexander Armstrong is joined by co-host Desiree Burch for the quiz where contestants try to score as few points as possible by coming up with the answers no-one else could think of. [S] [HD]</desc>
		<category>Show / Game show</category>
		<episode-num system="xmltv_ns">..44/47</episode-num>
	</programme>
```
image

### DIFF
--- a/src/iptvsimple/data/EpgEntry.cpp
+++ b/src/iptvsimple/data/EpgEntry.cpp
@@ -360,13 +360,19 @@ bool EpgEntry::ParseXmltvNsEpisodeNumberInfo(const std::string& episodeNumberStr
 
     if (!episodePartString.empty())
     {
-      int totalNumberOfParts;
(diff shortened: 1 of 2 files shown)


phunkyfish (changes requested):
Thanks, comment looks good. 

One small request to split the code change from the changelog and version. For iptvsimple we add a separate comment for the version bump. 

Eg: https://github.com/kodi-pvr/pvr.iptvsimple/commit/76b7ea9dd6b1bb2b0c7ed41a3ade1f936da307d6

phunkyfish (approved):

PR #927, opened 2024-11-21
Add mono fsListType parsing

am-silex:
Added format for catchup link using by my IPTV provider

### DIFF
--- a/src/iptvsimple/data/Channel.cpp
+++ b/src/iptvsimple/data/Channel.cpp
@@ -499,6 +499,8 @@ bool Channel::GenerateFlussonicCatchupSource(const std::string& url)
       {
         if (fsListType == "index")
           m_catchupSource = fsHost + "/" + fsChannelId + "/timeshift_rel-{offset:1}.m3u8" + fsUrlAppend;
+        else if (fsListType == "mono")
+          m_catchupSource = fsHost + "/" + fsChannelId + "/mono-{utc}-{duration}.m3u8" + fsUrlAppend;
         else
           m_catchupSource = fsHost + "/" + fsChannelId + "/" + fsListType + "-timeshift_rel-{offset:1}.m3u8" + fsUrlAppend;
       }


src/iptvsimple/data/Channel.cpp
```diff
@@ -499,6 +499,8 @@ bool Channel::GenerateFlussonicCatchupSource(const std::string& url)
       {
         if (fsListType == "index")
           m_catchupSource = fsHost + "/" + fsChannelId + "/timeshift_rel-{offset:1}.m3u8" + fsUrlAppend;
+        else if (fsListType == "mono")
```
phunkyfish, line 502: This won’t work I’m afraid as you’re forcing all `mono` list types to use your format, and for any `mono` list type they will never reach the following `else`.

The only way to do this would be to add a new catch format type so that you only test your condition and not any other `mono` types.

Plus, it would always fail for your format using an `index` list type.

phunkyfish (changes requested):

phunkyfish:
Can you add an example M3U entry (feel free to obfuscate any URLs).

am-silex:
Yes, I'm redirecting all _mono_ type to my catch as it's done with index.

Here is the original m3u link as in provider's playlist (obfuscated) 
http://iptvserver.net:1234/ch465/mono.m3u8?token=secret - LiveTV

After requesting actual data it turns into 
http://iptvserver.net:1234/ch465/2024/11/09/20/29/51-05005.ts?token=secret (I believe as it's done on any other streaming services - to get next 10-30 seconds of playback)

Catchup stream looks like this
http://iptvserver.net:1234/ch465/tracks-v1a1/index-1732173780-4500.m3u8?token=secret
And actual data after that
http://iptvserver.net:1234/ch465/tracks-v1a1/dvr-2024/11/21/07/23/00-06000.ts?token=secret

phunkyfish:
Sorry, I’m not looking for the M3U link for a stream. It’s a full M3U entry from the M3U file you load into iptvsimple that I’d like to see.

phunkyfish:
> Yes, I'm redirecting all _mono_ type to my catch as it's done with index.

Yes, I understand that, but this changes behaviour for every other iptvsimple user who has a mono stream. We can’t do this I’m afraid. It may solve your issue but will create other problems.

am-silex:
M3U playlist consists of links like that
http://iptvserver.net:1234/ch1/mono.m3u8?token=secret
http://iptvserver.net:1234/ch2/mono.m3u8?token=secret
and so on

phunkyfish:
So those are standard mono links. They are not different in any way and your change would modify existing behaviour for all users. The only way to do this cleanly should be to add a new catchup format type.

am-silex:
I'm not in any way an expert in streaming services, but when I've switched to Android TV box, almost any IPTV apps could open these links and detect right catchup scheme. Just works.
Frankly, I'm tired of re-compiling and tinkering with the soft only to be able to watch some channels.
Good luck!

phunkyfish:
Don’t give up so quickly, there might be a way to do what you need. We could add an advanced option, so you can toggle on the behaviour that you require in the addon settings. Would that work?

phunkyfish:
I created a PR adding an option to change from relative timeshift to utc and duration for Flussonic streams: https://github.com/kodi-pvr/pvr.iptvsimple/pull/960

Can you try the testbuilds: https://jenkins.kodi.tv/blue/organizations/jenkins/kodi-pvr%2Fpvr.iptvsimple/detail/PR-960/1/artifacts

PR #950, opened 2025-01-31
New feature web parser with pattern

alpgul:
## Description
Implement WebStreamExtractor to support dynamic URL extraction from web sources in playlist loading. This includes:
- New WebStreamExtractor utility class for extracting stream URLs
- Updated PlaylistLoader to support '@' prefixed URLs with optional web pattern
- Added CMakeLists.txt to include new source files
- Added Windows build script for easier compilation

## Missing Features
- Missing header support for web requests(CFile.OpenFile function doesn't support web header. what alternatives can be used?)
- In-depth web crawling is currently not supported.

## Example Usage
- usage with default pattern: 
```
#EXTINF:-1 tvg-name="Test" , Test
@http://127.0.0.1:3000/index.html
``` 
- usage with custom pattern: 
```
#EXTINF:-1 tvg-name="Test1", Test1
#WEBPROP:web-regex="(https?://[^\"]+\.m3u8)"
@http://127.0.0.1:3000/index.html
``` 
URL must be specified in parentheses.


phunkyfish:
Could you provide some examples? Preferably real world use cases.

The purpose of the feature is not clear to me from the PR description.

alpgul:
It is used to search for media URLs in HTML pages using a regex pattern.
Example usage: `
Define an m3u8 link in an HTML page. Then, define a regex to find the media URL link, and add the html link and regex as shown in the example usage above.`
Example Html Page:
```
<html>
...
url:"https://localhost:3000/index.m3u8"
...
</html>
```
Example Regex:`#WEBREGEX:url:"(https?://[^"]+\.m3u8)"`

The current issue is that I can't send requests using custom headers. Which library should I use that works on all platforms? I'm considering using CURL, but does it work on all platforms?

alpgul:
- Added support for headers in web URL requests.

## Example Usage

```
#EXTM3U
#EXTINF:-1 tvg-name="Test1", Test1
#WEBPROP:web-regex="([^"]+\.mp4)"
#WEBPROP:web-headers=user-agent:Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/132.0.0.0 Safari/537.36&referer:https://google.com/
@https://test-videos.co.uk/bigbuckbunny/mp4-h264

```

phunkyfish:
How do you ensure only one URL is selected, and also ensure it is the correct one?

phunkyfish:
Also note that new features must be tested and merged on the Piers branch before being considered for the Omega branch.

build-windows.bat
```diff
@@ -0,0 +1,68 @@
...
+    -DADDONS_DEFINITION_DIR="%CD%\definition" ^
+    -DADDON_SRC_PREFIX=../.. ^
+    -DCMAKE_INSTALL_PREFIX=../../kodi/addons ^
+    -DPACKAGE_ZIP=1 ^
+    ../../kodi/cmake/addons
+
+:: Build the project
+cmake --build . --config %CONFIGURATION% --target %app_id%
+
+pause
```
phunkyfish: New line missing

src/iptvsimple/PlaylistLoader.cpp
```diff
@@ -221,6 +224,13 @@ bool PlaylistLoader::LoadPlayList()
     }
     else if (line[0] != '#')
     {
+      if (line[0] == '@')
+      {
+        std::string extractedUrl = WebStreamExtractor::ExtractStreamUrl(line.substr(1), webPattern,isMediaEntry);
```
phunkyfish: space missing after `,`

src/iptvsimple/PlaylistLoader.cpp
```diff
@@ -221,6 +224,13 @@ bool PlaylistLoader::LoadPlayList()
...
+        Logger::Log(LEVEL_DEBUG, "%s - Extracted URL: '%s'", __FUNCTION__, extractedUrl.c_str());
+        line = extractedUrl;
+      }
```
phunkyfish, line 233: What if no valid URL is found?

Really what should have is that is whole M3U entry should be discarded. Or do you just use the default?

src/iptvsimple/utilities/WebStreamExtractor.cpp
```diff
@@ -0,0 +1,93 @@
...
+  for (const auto& pattern : patterns)
+  {
+    std::smatch match;
+    if (std::regex_search(content, match, pattern))
+    {
+      return match[2].str();
+    }
+  }
+  return "";
+}
```
phunkyfish: Missing newline

src/iptvsimple/utilities/WebStreamExtractor.h
```diff
@@ -0,0 +1,21 @@
...
+public:
+  static std::string ExtractStreamUrl(const std::string& webUrl, const std::string& pattern = "", bool isMediaEntry = false);
+
+private:
+  static std::string ExtractByPattern(const std::string& content, const std::string& customPattern, bool isMediaEntry);
+  static std::string DefaultFindUrl(const std::string& content, bool isMediaEntry);
+};
+
+} // namespace utilities
+} // namespace iptvsimple
```
phunkyfish: Missing newline

src/iptvsimple/PlaylistLoader.cpp
```diff
@@ -632,7 +632,7 @@ std::string PlaylistLoader::ReadMarkerValue(const std::string& line,
       size_t markerEnd;
       if (isCheckDelimiters)
       {
-        if ((marker == M3U_GROUP_MARKER || marker == WEB_REGEX_MARKER) && line[markerStart] != '"')
+        if ( marker == WEB_REGEX_MARKER ||(marker == M3U_GROUP_MARKER && line[markerStart] != '"'))
```
phunkyfish: Remove space after first `(`. Add space after `||`

src/iptvsimple/utilities/WebStreamExtractor.cpp
```diff
@@ -47,9 +47,9 @@ std::string WebStreamExtractor::ExtractByPattern(const std::string& content, con
     {
       std::regex pattern(customPattern);
       std::smatch match;
-      if (std::regex_search(content, match, pattern) && match.size() > 1)
-      {
-        return match[pattern.mark_count()].str();
+      if (std::regex_search(content, match, pattern) && match.size() > 0)
+      {
```
phunkyfish: Remove tab. Braces are not required as only a single statement in the `if`

src/iptvsimple/utilities/STRING.cpp
```diff
@@ -0,0 +1,10 @@
+#include "STRING.h"
+
+using namespace iptvsimple;
+using namespace utilities;
+
+int STRING::ToInt32(std::string_view str, int fallback) {
+     int result = fallback;
+  std::from_chars(str.data(), str.data() + str.size(), result);
+  return result;
+}
```
phunkyfish: We probably don't need a sepzarate function for this. Also, no filenames will all CAPS please.

phunkyfish (changes requested):
Just a light first review. I'd suggest collapsing all commits into a single one, as many of the commits change the previous one, making it difficult to review.

Also, many of the files are missing a newline at the end of the file.

phunkyfish:
`build-windows.bat` should be a separate commit.

src/iptvsimple/PlaylistLoader.cpp
alpgul: If the process fails, an empty string is returned. However, this process is now working within "GetChannelStreamProperties".

src/iptvsimple/PlaylistLoader.cpp
alpgul: This change has been reverted.

src/iptvsimple/utilities/STRING.cpp
alpgul: String, base64, and curlutils codes have been taken from the "inputstream.adaptive" project. I have removed the use of STRING.

src/iptvsimple/utilities/WebStreamExtractor.cpp
alpgul: I removed the braces.

alpgul:
> How do you ensure only one URL is selected, and also ensure it is the correct one?

Initially, the URL fetching process was in the playlist loader. However, now that it has been moved to "GetChannelStreamProperties" there should be no issues.



> `build-windows.bat` should be a separate commit.

It has been moved to a separate commit.

I merged the all commits.

depends/common/zlib/01-build-static.patch
```diff
@@ -1,36 +1,38 @@
...
  endif()
- 
+
  if(NOT SKIP_INSTALL_LIBRARIES AND NOT SKIP_INSTALL_ALL )
 -    install(TARGETS zlib zlibstatic
 +    install(TARGETS zlib
          RUNTIME DESTINATION "${INSTALL_BIN_DIR}"
          ARCHIVE DESTINATION "${INSTALL_LIB_DIR}"
-         LIBRARY DESTINATION "${INSTALL_LIB_DIR}" )
+         LIBRARY DESTINATION "${INSTALL_LIB_DIR}" )
```
phunkyfish: newline missing

depends/common/zlib/03-install-pkgconfig-in-lib.patch
```diff
@@ -1,12 +1,11 @@
...
  set(INSTALL_LIB_DIR "${CMAKE_INSTALL_PREFIX}/lib" CACHE PATH "Installation directory for libraries")
  set(INSTALL_INC_DIR "${CMAKE_INSTALL_PREFIX}/include" CACHE PATH "Installation directory for headers")
  set(INSTALL_MAN_DIR "${CMAKE_INSTALL_PREFIX}/share/man" CACHE PATH "Installation directory for manual pages")
 -set(INSTALL_PKGCONFIG_DIR "${CMAKE_INSTALL_PREFIX}/share/pkgconfig" CACHE PATH "Installation directory for pkgconfig (.pc) files")
 +set(INSTALL_PKGCONFIG_DIR "${CMAKE_INSTALL_PREFIX}/lib/pkgconfig" CACHE PATH "Installation directory for pkgconfig (.pc) files")
  
  include(CheckTypeSize)
- include(CheckFunctionExists)
-
+ include(CheckFunctionExists)
```
phunkyfish: missing newline

depends/common/zlib/zlib.sha256
```diff
@@ -1 +1 @@
-d14c38e313afc35a9a8760dadf26042f51ea0f5d154b0630a31da0540107fb98
\ No newline at end of file
+38ef96b8dfe510d42707d9c781877914792541133e1870841463bfa73f883e32
```
phunkyfish: missing newline

src/IptvSimple.cpp
```diff
@@ -239,6 +239,7 @@ PVR_ERROR IptvSimple::GetChannelStreamProperties(const kodi::addon::PVRChannel&
     else
       streamURL = m_catchupController.ProcessStreamUrl(m_currentChannel);
 
+    streamURL = StreamUtils::WebStreamExtractor(streamURL,m_currentChannel);
```
phunkyfish: missing space after comma

src/IptvSimple.cpp
```diff
@@ -417,6 +418,7 @@ PVR_ERROR IptvSimple::GetRecordingStreamProperties(const kodi::addon::PVRRecordi
 
   if (!mediaEntry.GetMediaEntryId().empty() && !url.empty())
   {
+    url = StreamUtils::WebStreamExtractor(url,mediaEntry);
```
phunkyfish: missing space after comma

src/iptvsimple/PlaylistLoader.cpp
phunkyfish: Would it not be better to just return the string without the `@` or just the URL if there no `@`. Otherwise we would break existing flows. Or do I have that wrong?

src/iptvsimple/utilities/Base64.cpp
```diff
@@ -0,0 +1,248 @@
...
+using namespace iptvsimple;
+using namespace utilities;
+namespace
+{
+constexpr char PADDING{'='};
+constexpr std::string_view CHARACTERS{"ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+                                      "abcdefghijklmnopqrstuvwxyz"
+                                      "0123456789+/"};
+// clang-format off
+constexpr unsigned char BASE64_TABLE[] = {
```
phunkyfish: What about using https://github.com/azawadzki/base-n instead as a dependency? It support `base64` out of the box. Then we don't need to carry this extra code.

As far as I can tell it's standard C++ so should be cross platform.

src/iptvsimple/utilities/CurlUtils.cpp
```diff
@@ -0,0 +1,178 @@
+#include "CurlUtils.h"
+
+using namespace iptvsimple;
+using namespace iptvsimple::utilities;
+
+namespace iptvsimple {
+namespace utilities {
+
+CUrl::CUrl(std::string_view url) {
```
phunkyfish: `{` on next line. Here and throughout all source files and headers. Even the namespaces.

https://github.com/xbmc/xbmc/blob/master/docs/CODE_GUIDELINES.md

src/iptvsimple/utilities/CurlUtils.cpp
```diff
@@ -0,0 +1,178 @@
...
+bool CUrl::IsEOF() {
+    return m_file.AtEnd(); 
+}
+
+bool DownloadFile(std::string_view url,
+                              const std::map<std::string, std::string>& reqHeaders,
+                              const std::vector<std::string>& respHeaders,
+                              HTTPResponse& resp) {
+    if (url.empty())
+        return false;
```
phunkyfish: 2 spaces indent, not 4, in all files

src/iptvsimple/utilities/CurlUtils.cpp
```diff
@@ -0,0 +1,178 @@
...
+  // to flag this (also implies chunked with HTTP1)
+  return contentLengthStr.empty() || transferEncodingStr.find("chunked") != std::string::npos;
+}
+
+bool CUrl::IsEOF() {
+    return m_file.AtEnd(); 
+}
+
+bool DownloadFile(std::string_view url,
+                              const std::map<std::string, std::string>& reqHeaders,
```
phunkyfish: See coding rules for indentation of functions

phunkyfish (changes requested):
Ok, review round 2.

Please take note https://github.com/xbmc/xbmc/blob/master/docs/CODE_GUIDELINES.md

src/iptvsimple/PlaylistLoader.cpp
alpgul: If we return the URL, `inputstream.adaptive`might send malicious requests to server, such as header checks. In this case, the server could detect and block you. That's why I return an empty string instead. Here, the use of "@" is only to check whether it is a web URL. If "@" is found, the code returns the URL without the "@" using line.substr(1) and continues normal behavior. This ensures the server doesn’t detect suspicious activity while maintaining the intended functionality.

src/iptvsimple/utilities/Base64.cpp
alpgul: I pulled this code from` inputstream.adaptive`. I chose this to minimize the risk of errors. If `base-n` is more performant, can we first adapt the changes in `inputstream.adaptive` and then apply them here as well?
[Reference Code Link](https://github.com/xbmc/inputstream.adaptive/blob/Piers/src/utils/Base64Utils.cpp)

alpgul:
> Ok, review round 2.
> 
> Please take note https://github.com/xbmc/xbmc/blob/master/docs/CODE_GUIDELINES.md

https://github.com/xbmc/xbmc/blob/master/.clang-format

The `.clang-format` file is not working properly because when I apply automatic formatting, the content of the code changes completely. In other words, the formatting file is incompatible with the previous code style, and I have to manually fix the code every time. If you have one, could you share the `.clang-format` file you are using?

src/iptvsimple/utilities/WebStreamExtractor.cpp
```diff
@@ -0,0 +1,122 @@
+#include "WebStreamExtractor.h"
+
+using namespace iptvsimple::utilities;
+
+std::string WebStreamExtractor::ExtractStreamUrl(const std::string& webUrl,
```
alpgul, line 5: This code explained in [this link](https://github.com/kodi-pvr/pvr.iptvsimple/pull/950/commits/a0627e9638d232cf7733663e0b05d873ed6ed7ba#r1947544345)

src/iptvsimple/utilities/CurlUtils.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+using namespace iptvsimple;
+using namespace iptvsimple::utilities;
+
+namespace iptvsimple
+{
+namespace utilities
+{
+
+CUrl::CUrl(std::string_view url)
```
alpgul, line 11: This code has been adapted from the following [link](https://github.com/xbmc/inputstream.adaptive/blob/Piers/src/utils/CurlUtils.cpp).

src/iptvsimple/PlaylistLoader.cpp
phunkyfish: Ok, that logic makes sense.

src/iptvsimple/utilities/Base64.cpp
phunkyfish: I would probably take a different view on this. We should try the non custom code approach in iptvsimple and if that proves to work correctly then propose that dependency change for adaptive but not the other way around.

phunkyfish:
> > Ok, review round 2.
> > Please take note https://github.com/xbmc/xbmc/blob/master/docs/CODE_GUIDELINES.md
> 
> https://github.com/xbmc/xbmc/blob/master/.clang-format
> 
> The `.clang-format` file is not working properly because when I apply automatic formatting, the content of the code changes completely. In other words, the formatting file is incompatible with the previous code style, and I have to manually fix the code every time. If you have one, could you share the `.clang-format` file you are using?

That clang format file is for xbmc, we don't follow all the conventions in addons just some of the more prominent ones to make code easier to read. I understand that doesn't make it black and white for  contributors but it's where we are currently.

src/iptvsimple/utilities/CurlUtils.cpp
phunkyfish: Ok, but we don't use most of the functions in this file. From a brief look we only use Read, AddHeaders, and open in addition to one form of the constructor. Can we remove the unused functions?

phunkyfish (changes requested):

alpgul:
Removed Unused Methods in CurlUtils and Base64

alpgul:
I made the code changes for the Omega branch by reviewing the changes in the Piers branch.

phunkyfish:
You need to update the conflicting files as there have been a few releases since

alpgul:
I’ve applied the updates.

phunkyfish (approved):

phunkyfish:
v21.11.0
- Add support for web stream extraction from HTML content
- Add support for custom regex patterns for web stream extraction

PR #955, opened 2025-02-06
Add GitHub Actions workflow for WebOS IPTV Simple PVR addon build

alpgul:
**Added Workflow:**

-   Added a new workflow to support building addons for webOS.
-   Implemented cache management to optimize performance.
    -   Cache covers 2GB of storage space.
-   Improved application runtime:
    -   Without cache: ~30 minutes.
    -   With cache: ~1.5 minutes.

### DIFF
--- a/.github/workflows/webos-build.yml
+++ b/.github/workflows/webos-build.yml
@@ -87,6 +87,7 @@ jobs:
       - name: Checkout add-on repo
         uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


phunkyfish:
Really WebOS should be build as part of Jenkins and not as a separate job.

alpgul:
The Jenkins artifact outputs can also be added to the release notes or asserts so that users can try them out.

PR #959, opened 2025-02-16
Create devcontainer.json

maxou3535:


phunkyfish:
I don't understand why this change needs to be committed to the repo. We don't add build platform specific files to the binary addon repos.

PR #962, opened 2025-02-21
Improve channel and EPG matching with Sørensen-Dice similarity algorithm

alpgul:
- Add StringComparer utility to calculate string similarity
- Enhance channel and EPG matching to use fuzzy string comparison
- Update channel and EPG finding methods to support more flexible matching
- Modify EPG entry description to include channel ID

### DIFF
--- a/src/iptvsimple/data/EpgEntry.cpp
+++ b/src/iptvsimple/data/EpgEntry.cpp
@@ -214,6 +214,7 @@ bool EpgEntry::UpdateFrom(const xml_node& programmeNode, const std::string& id,
     return false;
 
   m_broadcastId = static_cast<int>(programmeStart);
+  m_plot = id + "\n" + GetNodeValue(programmeNode, "desc");
   m_channelId = std::atoi(id.c_str());
   m_genreType = 0;
   m_genreSubType = 0;
@@ -227,7 +228,6 @@ bool EpgEntry::UpdateFrom(const xml_node& programmeNode, const std::string& id,
   m_seasonNumber = EPG_TAG_INVALID_SERIES_EPISODE;
 
   m_title = GetNodeValue(programmeNode, "title");
-  m_plot = GetNodeValue(programmeNode, "desc");
   m_episodeName = GetNodeValue(programmeNode, "sub-title");
 
   m_genreString = GetJoinedNodeValues(programmeNode, "category");
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,7 +45,8 @@ set(IPTV_SOURCES src/addon.cpp
                  src/iptvsimple/utilities/Logger.cpp
                  src/iptvsimple/utilities/SettingsMigration.cpp
                  src/iptvsimple/utilities/StreamUtils.cpp
-                 src/iptvsimple/utilities/WebUtils.cpp)
+                 src/iptvsimple/utilities/WebUtils.cpp
+                 src/iptvsimple/utilities/StringComparer.cpp)
 
 set(IPTV_HEADERS src/addon.h
                  src/IptvSimple.h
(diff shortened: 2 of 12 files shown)


alpgul (commented):
I didn't add the algorithm to the `MergeEpgDataIntoMedia` method in pvr.iptvsimple\src\iptvsimple\Epg.cpp because it causes performance loss. It also needs to be debugged as it returns "-1" on the first run, contains a bug

alpgul (commented):
[const MediaEntry* Media::FindMediaEntry(const std::string& id, const std::string& displayName) const](https://github.com/alpgul/pvr.iptvsimple/blob/48abda6db694dfc2b28411fcb6cb3ffa7f71bffb/src/iptvsimple/Media.cpp#L158)
[MergeEpgDataIntoMedia](https://github.com/alpgul/pvr.iptvsimple/blob/48abda6db694dfc2b28411fcb6cb3ffa7f71bffb/src/iptvsimple/Epg.cpp#L680)
For optimization purposes, the execution of `FindMediaEntry` and `MergeEpgDataIntoMedia` methods should be prevented if the media is in recordings. Additionally, an isMedia attribute value should be defined in the channel definition of epg.xml, thus preventing unnecessary performance loss.

src/iptvsimple/data/EpgEntry.cpp
```diff
@@ -214,6 +214,7 @@ bool EpgEntry::UpdateFrom(const xml_node& programmeNode, const std::string& id,
     return false;
 
   m_broadcastId = static_cast<int>(programmeStart);
+  m_plot = id + "\n" + GetNodeValue(programmeNode, "desc");
```
phunkyfish, line 217: Why prepend the id onto the plot?

phunkyfish:
Many channel ids are short in nature.

for example: BBC1, BBC2, BBC3 etc.

if you have these channels but only BBC4 in the xmltv data what happens?

Matches for channel ids are supposed to be the exact by there nature as IDs.

I think for this feature we should put it behind a setting somewhere a user can switch it on if they choose to do so. WDYT?

alpgul (commented):
A minimum similarity percentage can be added in settings. I included this code `m_plot = id + "\n" + GetNodeValue(programmeNode, "desc"); `as an example to help in such cases. This is the actual EPG channel ID value being added. I added it to the plot since it doesn't look visually disruptive. Additional usable tags from the EPG screen can be added there if available.

phunkyfish:
> A minimum similarity percentage can be added in settings. I included this code `m_plot = id + "\n" + GetNodeValue(programmeNode, "desc"); `as an example to help in such cases. This is the actual EPG channel ID value being added. I added it to the plot since it doesn't look visually disruptive. Additional usable tags from the EPG screen can be added there if available.

Ok, but this should only be for testing and not in the final PR. I suggest a Debug log statement that can be grepped for to find false positives.

The default similarity setting should be 100% match, then users or addons can set this to another value to enable fuzzy matching. You could also supply some labelled values instead of users trying to set a magic number.

alpgul (commented):
Add EPG channel name match threshold setting

- Introduce new setting to control channel name matching similarity
- Add configuration for minimum similarity percentage when matching EPG channels
- Update settings XML, language strings, and implementation to support new threshold

alpgul (commented):
I've set it to 100%. I also set the increment amount to 5. the similarity between `BBC1` and `BBC4` comes out as 75%, it shows good performance at 80% threshold

src/iptvsimple/data/EpgEntry.cpp
alpgul: Is there a more suitable tag where I can display the EPG Channel ID value in the EPG View? This would help users identify any incorrect matches

PR #966, opened 2025-03-27
Fix and add support to inputstream.adaptive.common_headers

### LINKED ISSUE #925: [bug] Simple IPTV Client Overriding User-Agent in InputStream Adaptive common_headers
The Simple IPTV Client is overriding the User-Agent specified in InputStream Adaptive’s common_headers property. This behavior prevents the custom User-Agent set in InputStream Adaptive from being sent for adaptive streams, and instead, the UA defined in Advanced Settings of the Simple IPTV Client is used.

### Steps to Reproduce:

1. Configure InputStream Adaptive to use a custom User-Agent (UA) by adding it to the common_headers property as outlined in [InputStream Adaptive Documentation](https://github.com/xbmc/inputstream.adaptive/wiki/Integration#inputstreamadaptivecommon_headers-).
2. Set up UA in SIPTV's advanced settings 
3. Play a stream
4. Inspect the network traffic or check the Kodi log to verify which User-Agent is being sent.

### Expected Behavior:

- When a custom User-Agent is set in InputStream Adaptive’s common_headers property, the Simple IPTV Client should not override it and should send the specified UA for the streams.
- Simple IPTV Client should only override the User-Agent when no UA is provided in InputStream Adaptive's properties, using its own config in that case.

### Actual Behaviour:

The Simple IPTV Client is overriding the User-Agent specified in InputStream Adaptive's common_headers property, even when a custom UA is defined. This causes the Simple IPTV Client to send the User-Agent defined in its Advanced Settings, which can cause issues.

### Additional Information:
- **Kodi Version**: 22.r1532.g65703fe-1
- **ISA Version**: 22.1.9
- **SIPTV Version**: 22.4.1

- The common_headers property in InputStream Adaptive was introduced in this [PR](https://github.com/xbmc/inputstream.adaptive/pull/1687).
- After reviewing the code in the [SIPTV source file](https://github.com/kodi-pvr/pvr.iptvsimple/blob/Piers/src/iptvsimple/utilities/StreamUtils.cpp), it is clear that the current implementation only handles manifest headers and stream headers.

CastagnaIT:
stream_headers and manifest_headers can replace common_headers values since they have higher priority

so when a stream provide a common_headers only
you dont have to set stream_headers, manifest_headers, otherwise will override user customizations

since common_headers (supported starting from Piers branch) can be used to set the headers
globally, we make use of this prop instead of the other

fixes #925

### DIFF
--- a/pvr.iptvsimple/addon.xml.in
+++ b/pvr.iptvsimple/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.iptvsimple"
-  version="22.5.0"
+  version="22.5.1"
   name="IPTV Simple Client"
   provider-name="nightik and Ross Nicholson">
   <requires>@ADDON_DEPENDS@
--- a/pvr.iptvsimple/changelog.txt
+++ b/pvr.iptvsimple/changelog.txt
@@ -1,3 +1,7 @@
+v22.5.1
+- Add support to inputstream.adaptive.common_headers to set headers
+- Fix an issue that was replacing user headers set with inputstream.adaptive.common_headers
+
 v22.5.0
 - Add support for web stream extraction from HTML content
 - Add support for custom regex patterns for web stream extraction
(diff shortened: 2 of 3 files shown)


CastagnaIT:
@phunkyfish can you review?

phunkyfish (approved):

phunkyfish:
Change looks good. But you'll need to update the `addon.xml.in` and `changelog.txt` as they now have conflicts

CastagnaIT:
done rebased and fix conflicts

PR #968, opened 2025-04-12
Add Windows build script for PVR IPTV Simple Add-on

alpgul:
This pull request introduces support for building the `pvr.iptvsimple` addon on Windows. Below are the key changes included in this PR:

-   Added necessary configurations and scripts for Windows build.
-   Incorporated configurations inspired by the [Azure Pipelines configuration](https://github.com/kodi-pvr/pvr.iptvsimple/blob/Piers/azure-pipelines.yml) to ensure seamless integration and compatibility.

Testing:

-   Verified successful builds on Windows 10 x64.

### DIFF
--- a/build-windows.bat
+++ b/build-windows.bat
@@ -1,0 +1,70 @@
+@echo off
+setlocal enabledelayedexpansion
+
+:: Variables
+set "app_id=pvr.iptvsimple"
+set "GENERATOR=Visual Studio 17 2022"
+set "CONFIGURATION=Release"
+
+:: Get user architecture selection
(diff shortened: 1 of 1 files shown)


phunkyfish:
Hi, we likely will not merge this. The build system is cmake so we don't ship platform specific build files by default.

The one exception to this is MacOS as it's a development environment with lesser support than windows/Linux. Without the script its not possible to build debug addon's that can be run with Kodi. Hopefully this will be fixed in cmake in the future.

alpgul:
Got it, thanks!

PR #978, opened 2025-06-15
Patch zlib build for cmake 4

emveepee:
Upgrade the minimum version to 3.5 as required by cmake 4.  Only tested generate and build

### DIFF
--- a/depends/common/zlib/04-cmake4-minimum.patch
+++ b/depends/common/zlib/04-cmake4-minimum.patch
@@ -1,0 +1,8 @@
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -1,4 +1,4 @@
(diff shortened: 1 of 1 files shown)


fuzzard:
Why not just bump zlib to 1.3 or 1.3.1?

emveepee:
> Why not just bump zlib to 1.3 or 1.3.1?

Primarily because of full testing which is out of scope for me.  If @phunkyfish prefers I could submit depends/common/zilib from Piers.

phunkyfish:
That might be easier

emveepee:
> That might be easier

If you mean bring in Piers zlib it isn't easier but I can do that if you want.

phunkyfish:
I meant from a maintenance perspective, at least then you know it's tested.

PR #1035, opened 2026-08-24
Add CatchupMode::RESOLVER: dynamic per-request catch-up stream property overrides

hkoeck:
Stacked on #1034 — this branch includes those two commits. The diff below
will shrink to just the RESOLVER commit once #1034 merges.

### Problem
Some providers' catch-up content needs a per-request resolved value that
can't be baked statically into the M3U — e.g. a DRM license/manifest that
differs from the channel's live-stream value and is only resolvable via
a live backend call at playback time. There's currently no way to plug
dynamic property resolution into inputstream.adaptive catch-up playback;
`SetCatchupInputStreamProperties()` already does something similar, but
only for inputstream.ffmpegdirect.

### Changes
New opt-in catchup mode, `catchup="resolver"` in the M3U, matching the
existing `catchup="vod"/"flussonic"/"xc"` convention:

- `CatchupMode::RESOLVER` appended after `VOD` in the enum (not
  inserted) to keep existing installs' persisted integer settings
  stable.
- For RESOLVER-mode channels, `GetEPGTagStreamProperties()` fetches the
  catchup-source URL (a resolver endpoint, not a directly playable URL)
  and parses the response as newline-separated `key=value` pairs. The
  special key `streamUrl` becomes the actual play URL; every other key
  is merged into the stream properties map as-is — so a resolver can
  override `license_key`, `license_type`, custom headers, or anything
  else, not just one hardcoded field. Plain text, no JSON dependency
  added.
- Wired through the same places every other catchup mode is
  (`GetCatchupModeText`, `IsValidTimeshiftingCatchupSource`,
  `ConfigureCatchupMode`, M3U parsing in `PlaylistLoader.cpp`).

### Test plan
- Built clean against Kodi 22 (Piers), no new warnings.
- Deployed against a real IPTV provider requiring per-request DRM
  license resolution for catch-up; verified end-to-end: live-only
  license correctly rejected for catch-up playback with the static
  M3U license, RESOLVER mode's fetched override plays back correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,6 +244,7 @@ Addon settings for catchup:
     - `Shift (SIPTV)` - Append the standard SIPTV catchup string to the channel URL.
     - `Flussonic` - Build a flussonic URL from the channel URL.
     - `Xtream codes` - Build an Xtream codes URL from the channel URL.
+    - `Resolver` (`catchup="resolver"` in the M3U) - For providers where the catch-up stream needs a per-request resolved value that can't be baked statically into the M3U (e.g. a DRM license/manifest that differs from the channel's live-stream value). The `catchup-source` is not a playable URL itself but a resolver endpoint: it is fetched at playback time and its response - plain text, newline-separated `key=value` pairs - is applied as stream property overrides instead of being played directly. The special key `streamUrl` supplies the actual manifest/stream URL to play; every other key (e.g. `license_key`, `license_type`, custom headers) is merged into the stream properties as-is.
 * **Override catchup for channels**: Set the scope for overriding the catchup mode. Options are:
     - `without catchup mode` - Only include channels with no catchup mode set (except legacy SIP `timeshift` catchup mode).
     - `with catchup mode` - Only include channels with catchup mode set (ignore those without a catchup mode).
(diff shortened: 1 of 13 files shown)


phunkyfish:
The comments in this PR make little sense at all.  Happy to review once they are cleaned up, or removed if they make no sense.

Example:

```
    RESOLVER // catchup-source is a resolver endpoint, fetched at playback
             // time and returning stream property overrides rather than a
             // directly playable URL - see FetchCatchupSourceOverrides() in
             // IptvSimple.cpp. Appended last to keep existing enum ordinals
             // (persisted in addon settings as a plain integer) stable.
```

This should be explained in the readme, but as a comment to an enum value is excessive

hkoeck:
Fair point, thanks for the honest feedback. I've gone through the whole PR and trimmed every comment like that down to a line or two - just the invariant/why, not the investigation log.

For context on the RESOLVER mode itself: some providers give live and catch-up content different DRM license/manifest values for the same channel, and the catch-up one can only be resolved with a live backend call at playback time rather than baked into the M3U statically. `SetCatchupInputStreamProperties()` already handles something similar for inputstream.ffmpegdirect, so this generalizes the same idea for inputstream.adaptive as an opt-in mode, without needing to patch the addon per-provider. That's now documented properly in the README (Catchup modes section + an M3U example) instead of living in a code comment.

Force-pushed the cleaned-up version - should be much easier to review now. Let me know if anything's still unclear or too terse in the other direction.

PR #1039, opened 2026-09-05
CatchupController: return the catchup stream key

heitbaum:
GetStreamKey built the catchup key and then discarded it, so catchup and live streams for a channel shared one key and the StreamType/MimeType lookup in StreamManager would return the wrong pairing.


eigendude:
@codex review

garbear (approved):
Fix does what the PR says.

PR #1041, opened 2026-09-10
[pvr.iptvsimple] Add support for PVR.Scan JSON-RPC call

Lunatixz:
// Enables Kodi's PVR.Scan JSON-RPC method to trigger a channel scan 
// on this addon. When called, sets the reload flag which the background 
// Process() loop picks up to re-fetch M3U playlist and EPG data. 
//
// Changes:
// - SetSupportsChannelScan(true) in GetCapabilities() 
// - Implement OpenDialogChannelScan() override
//
// PVR.Scan clientid parameter: optional, skips dialog if provided.

### DIFF
--- a/src/IptvSimple.cpp
+++ b/src/IptvSimple.cpp
@@ -111,6 +111,7 @@ PVR_ERROR IptvSimple::GetCapabilities(kodi::addon::PVRCapabilities& capabilities
   capabilities.SetSupportsDescrambleInfo(false);
   capabilities.SetSupportsRecordings(true);
   capabilities.SetSupportsRecordingsDelete(false);
+  capabilities.SetSupportsChannelScan(true);
 
   return PVR_ERROR_NO_ERROR;
 }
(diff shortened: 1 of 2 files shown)


src/IptvSimple.cpp
```diff
@@ -450,6 +451,16 @@ PVR_ERROR IptvSimple::StreamClosed()
   return PVR_ERROR_NO_ERROR;
 }
 
+/***************************************************************************
+ * Channel Scan
+ **************************************************************************/
+
+PVR_ERROR IptvSimple::OpenDialogChannelScan()
```
phunkyfish, line 458: We shouldn't use this override, instead `StartChannelScan` would fit and reflect what the intention is.

phunkyfish (changes requested):

src/IptvSimple.cpp
Lunatixz: The SDK only lists OpenDialogChannelScan(); no StartChannelScan method on CInstancePVRClient?

Lunatixz: StartChannelScan is a wrapper for OpenDialogChannelScan? they are one in the same, or am I missing something?

